PR #5, opened 2021-02-18
Make sure to download GDX files for tests in binary mode.

Loisel:
@Renato-Rodrigues experienced problems with the MD5 hash checking (on windows, I suppose). I forced *binary-mode* for the download, so that silently replaced line-endings should not come in the way.


Loisel:
Please test @Renato-Rodrigues

Renato-Rodrigues:
Thks @Loisel 
This solves the issue :)

PR #12, opened 2021-02-24
Add a test for compareScenarios.

Loisel:
this test runs atop the GDX test, to avoid having to produce the
reporting twice. Runtime of the build is now **9m 40s** which is substantial. I wonder what to do about this, since I think it is important to have a (rudimentary) test on `compareScenarios`. One possibility would be to perform this check only in the CI on github actions. Would require some "skip_if" clause.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '20623824'
+ValidationKey: '20737020'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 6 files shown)


Loisel:
maybe consider this a draft for now until we find a good solution

Loisel:
@giannou I pulled the newest changes and the test now completes successfully. From what I get from the REMIND meeting the PR can be merged.

giannou (approved):

PR #14, opened 2021-02-25
Disabled tests in RStudio 

giannou:
they still run if when you run lucode2::buildLibrary. corrected the line that you need to comment out in order to disable the tests if you are not using RStudio

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '20737020'
+ValidationKey: '20756813'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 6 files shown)


Loisel:
I guess this will not lead to the desired results, as `succeed()` seems not to return from the test function (as opposed to `fail()`). What should be used instead is `skip()` above the `test_that` call. This will be part of the PR #12

giannou:
> I guess this will not lead to the desired results, as `succeed()` seems not to return from the test function (as opposed to `fail()`). What should be used instead is `skip()` above the `test_that` call. This will be part of the PR #12

yeah I just noticed, I'll push changes. Do you want to merge your PR first?

PR #20, opened 2021-03-15
adjustment of reporting to new synfuel representation and FE prices

fschreyer:
This PR contains adjustment of reporting to new synfuel representation, the addition of a new FE price reporting and  minor bugfixes and cleaning up. 

Fast review would be great as we need it for looking at current results.


Loisel (approved):
Looks fine. I guess this will be tested thoroughly once the new output for ARIADNE will be produced.

Renato-Rodrigues:
@fschreyer As synfuel is added now as an alternative FE, the compare scenarios function also needs to be updated.

PR #25, opened 2021-03-19
some bugfixes and ariadne-related variable additions

fschreyer:
some bugfixes and ariadne-related variable additions

@giannou : Below, I hard-coded a path to a data file (pm_fe_nechem.cs4r) that I need (temporarily) for the new FE non-energy use variables. Could you please tell me where I can drop this file on the cluster such that the library automatically loads it. I will adapt the path before merging. 

PS: Ah and I just noticed. It might require a further corresponding bugfix with synfuels for demand-side emissions. I will check that and push again. Looking forward to emissions with ``+`` ;-)


R/reportFE.R
```diff
@@ -1163,10 +1176,91 @@ reportFE <- function(gdx,regionSubsetList=NULL,t=c(seq(2005,2060,5),seq(2070,211
...
+    
+    if ("DEU" %in% getRegions(vm_prodFe)) {
+      
+      # some initializations required for building library with dplyr operations below
+      encar <- data <- value <- value_subsectors <- SSP <- Value_NonEn <- encar <- region <- period <- NULL
+      
+      
+      
+      # read in FE industry non-energy use trajectories from industry subsectors run
+      df.fe_nechem <- read.csv("C:/work/scripts/pik-piam/misc/pm_fe_nechem.cs4r",
```
giannou, line 1192: Here's how you can access a data file from inside a package: Put the file in `inst/extdata` and access it with the command `system.file`. An example is found [here](https://github.com/pik-piam/mip/blob/605ccbb41e425f787916210fc959f058e8ddf49e/R/plotstyle.R#L43) for the file found [here](https://github.com/pik-piam/mip/blob/master/inst/extdata/plotstyle.csv)

giannou (changes requested):

giannou (approved):

PR #26, opened 2021-03-22
Additions to FE variables and compare scenario pdf for ariadne

fschreyer:
Additions to FE variables and compare scenario pdf for ariadne

* added further FE non-energy variables
* added SE trade to compare scenario pdf
In compare scenario, FE price plots put in if bracket such that it shows FE prices either with old or new reporting. Old Reporting should be removed at some point.


fschreyer:
Sry, I merged myself because I need it timely for the ariadne submission. In case of any issues, I am ready to hotfix.

PR #27, opened 2021-03-24
Refactoring and adapting SE demand variables to SE trade, quick fixes emissions

fschreyer:
Refactoring and adapting SE demand variables to SE trade, quick fixes emissions (temporary, will be harmonized/overhauled once new emissions reporting).


Renato-Rodrigues (approved):
I did not checked thoroughly, but I agree with the change. You can merge it from my point of view.

PR #43, opened 2021-05-11
Refactor the price calculation

Loisel:
- using new pm_SE/FEPrice and p_PEPrice variables from GDX.
- *mixed* prices on PE and SE level, e.g., for SE|Liquids are removed
- all vars follow the convention `Prices|<Energy Level>|<other specifiers>`
- FE vars follow the convention `Prices|<Energy Level>|<Sector>|...`
- Biomass debug variables (e.g., shiftfactor) are moved to `Internal|`
- Hydrogen based synthetic fuels are all labelled `|Hydrogen` (and not `|Synthetic (CCU)`)
- For all price signals in the variables there is an averaged variant. The same weights are used as for the non-averaged variables (automatically).
- I removed also ESD/ETS specific prices (e.g., `Price|Final Energy|Liquids|Transport|LDV|w/ costs for emissions|ESD (US$2005/GJ)`, as they were not clear to me (all prices include emission costs on FE level?) but maybe we want to keep them

This PR will break all pre-`pm_PE/SE/FEPrice` GDXs. On the upside, pre-`seliqXXX` and `segaXXX` if-clauses could be deleted. Exceptions for `stat_mod=simple` were deleted (fully trusting what's in the new price vars).


R/reportPrices.R
```diff
@@ -31,134 +31,164 @@
...
+      out,
+      setNames( dimSums(mselect(p_weights_FEprice_mkt * pm_FEPrice, all_enty1="fegat", emi_sectors = "trans"), dim=3.3, na.rm = T)*tdptwyr2dpgj,
+               "Price|Final Energy|Transport|Gases (US$2005/GJ)"))
+  }
+
+  out <- mbind(
+    out,
+    setNames( dimSums(mselect(p_weights_FEprice_mkt * pm_FEPrice, all_enty1="fedie", emi_sectors = "trans"), dim=3.3, na.rm = T)*tdptwyr2dpgj,
+             "Price|Final Energy|Transport|Liquids|HDV (US$2005/GJ)"),
+    setNames( dimSums(mselect(p_weights_FEprice_mkt * pm_FEPrice, all_enty1="fepet", emi_sectors = "trans"), dim=3.3, na.rm = T)*tdptwyr2dpgj,
```
fschreyer, line 90: Please change the name according to the new naming convention to ``Price|Final Energy|Transport|LDV|Liquids``. We now want to do all sectors first, then energy carriers, then energy sources or energy uses.

R/reportPrices.R
```diff
@@ -31,134 +31,164 @@
...
-  
-  
-  
-  
-  
-  
-  
-  
-  
+
```
fschreyer, line 191: I mean, I guess in the whole section below quite some variables like ``sebal`` and ``finenbal`` and stuff can be deleted in this old reporting part as they are not used anymore. We would need to clean this up another time maybe.

fschreyer: As far as I see it is only the "Internal" biomass prices, the energy services prices and the carbon prices which are in this section of the "old" reporting now, right?

fschreyer (changes requested):

fschreyer:
Btw, thanks a lot, Alois, for all this cleaning up of the price reporting so far!

R/reportPrices.R
Loisel: yeah I thought about this. The problem I see here is that these are essentially also different types of liquids. So in this sense it follow your logic already. It is also confusing, if you take HDV/LDV as mode type, that the same distinction is not made for other carriers (there is no `HDV|Electric` and the like).

Loisel: It also does not make sense to add `HDV|Electric` since the price signal is equivalent.

fschreyer (approved):

PR #50, opened 2021-06-14
Corrections in energy investment subcomponents by CB

giannou:


christophbertram (commented):
Most of this is actually pure cosmetics, which makes the real substantial changes a bit hard to see. Those are all in the block from lines 83-147 in reportEnergyInvestment.R, and relate to Hydrogen (l.114-115), Liquids (119-121), and the Supply total (l. 138-145).

giannou:
> Most of this is actually pure cosmetics, which makes the real substantial changes a bit hard to see. Those are all in the block from lines 83-147 in reportEnergyInvestment.R, and relate to Hydrogen (l.114-115), Liquids (119-121), and the Supply total (l. 138-145).

In the Files changed tab you can go to the gear icon and check the box that hides whitespace changes

PR #69, opened 2021-07-07
Annual milege added to EDGE-T reporting.

MariannaR:
this PR contains:

- annual mileage loaded from input data instead of using estimate
- removed useless lines in `reportFE` in `edge_esm` realization


Loisel (approved):
Nice. With this we have consistent mileage everywhere, no?

PR #70, opened 2021-07-08
bugfix variables that failed summation check and enable summation check again

fschreyer:
bugfix variables that failed summation check and enable summation check again

I built the library after enabling the summation checks so there should be no more issues when building.


giannou (approved):

giannou:
thanks Felix, great work!

PR #82, opened 2021-08-12
remove magpie dependency

pfuehrlich-pik:
magpie::read_all was the only function needed from the magpie package. The same function is now available as remind2::readAll, so the dependency on magpie is no longer needed. I removed GitHub actions, because there's only one test at the moment, and that could not be properly executed, because gdxrrw is not initialized correctly in GitHub actions.


pfuehrlich-pik:
I tested the only function that used read_all (remind2::readSupplycurveBio) with the new remind2::readAll on the cluster, and it still works fine. I also successfully ran the supplycurves REMIND output script.

PR #88, opened 2021-08-25
add final energy variables per carrier without bunkers and without non-energy use

fschreyer:


Renato-Rodrigues:
What is the default? Variables without any specific mention should include non-energy by default?
Ex: "FE|w/o Bunkers|Liquids (EJ/yr)" include non-energy?

If this is not the case, I don't think you need the "w/o Non-energy Use" term in "FE|w/o Bunkers|w/o Non-energy Use|Liquids (EJ/yr)" for example.
If the default is to include energy and non-energy by default, you can just disregard my question. 
I just wanted to clarify that.

Renato-Rodrigues (approved):

fschreyer:
Ye, by default FE includes non-energy and bunkers.

Issue #89, opened 2021-08-25
Values swapped between scenarios due to partly repeated names


giannou:
This code chunk seems to contain a bug: https://github.com/pik-piam/remind2/blob/c6380a4260a4c9fdae8afd68945e4df9da6305d6/R/compareScenarios.R#L165

When comparing 3 scenarios, 2 of which have identical scenario names, values are exchanged between the unique scenario and one of the two repeated ones.

cchrisgong:
Thanks for pointing this out. By values swapped, do you mean if the original scenarios are (A,A,B), you get A1, B1, A, instead of A1, A2, B?

cchrisgong:
Thanks for pointing this out. Just to make sure, as an example, if the original scenarios are [A,B,A], it becomes A1, A2, B1, correct?

giannou:
> Thanks for pointing this out. Just to make sure, as an example, if the original scenarios are [A,B,A], it becomes A1, A2, B1, correct?

yes

giannou:
fixed with #90

PR #90, opened 2021-08-25
bug fix for issue 89

cchrisgong:
now repeated scenario names are not swapped
i.e. original names = (A,B,A)
new names = (A1,B,A2)
(before [bugged version](https://github.com/pik-piam/remind2/issues/89) was A1, A2, B1)

### DIFF
--- a/R/compareScenarios.R
+++ b/R/compareScenarios.R
@@ -175,23 +175,22 @@ compareScenarios <- function(mif, hist,
     warning(paste0("There are repeated scenario names in comparison! Appending numerical value to repeated names"))
   }
   
-  # append numerical value to repeated names
-  mifScenarios_new <- c()
-  for(i in 1:length(uniqueScenarioName)){
-    unique_element = uniqueScenarioName[[i]]
-    k = 1
-    for(j in 1:length(mifScenarios)){
-      if (mifScenarios[[j]] == unique_element){
-        mifScenarios_new <-c(mifScenarios_new, paste0(unique_element, "_", k))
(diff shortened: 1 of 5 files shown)


R/compareScenarios.R
```diff
@@ -171,28 +171,28 @@ compareScenarios <- function(mif, hist,
...
+      dup_name = dup_names[[i]]
+      #number of duplicates
+      dupnum <- seq(1,length(mifScenarios[mifScenarios == dup_name]))
+      mifScenarios[mifScenarios == dup_name] <- paste0(mifScenarios[mifScenarios == dup_name], "_", dupnum)  
+    }
+    
+    mifScenarios_new <- mifScenarios
+    #rename scenario names
+    for(i in 1:length(mifScenarios)){
+      magclass::getNames(mifData[[i]],dim=1) <- mifScenarios_new[[i]]
```
giannou, line 192: why is `magclass::` used here? not needed

giannou (changes requested):

R/compareScenarios.R
cchrisgong: I used this "magclass::getNames" originally since many lines below also contain "magclass::getNames" (but some other times only getNames is used). I am not sure which is correct/needed. Should I only remove this one or all the subsequent ones also (those lines are not from me)?

giannou: feel free to remove only yours :)

cchrisgong: done :)

PR #97, opened 2021-08-31
include DTcoup power module in LCOE scripts

cchrisgong:


cchrisgong:
This fix is extremely similar to the last one, so no reviewers needed

PR #100, opened 2021-09-02
Refactor variables depending on emiTe

spahrm:
Inside REMIND several variable names will be refactored. This variables will then be depending on `emiTe` instead of `all_enty`.
By checking via `getSets` all variables, later depending on `emiTe` instead of `all_enty` , use `emiTe` and as a default and for backward compability the `all_enty` fallback is still implemented.

Since this is my first PR at PIK and I'm also not an R expert, please check the code carefully and tell me if I could improve the code or if I missed something critical.


LaviniaBaumstark (approved):

R/reportEmi.R
```diff
@@ -517,17 +517,25 @@ reportEmi <- function(gdx, output=NULL, regionSubsetList=NULL,t=c(seq(2005,2060,
...
-  out <- mbind(out,
-               setNames(
-               # vm_emiTeMkt is variable in REMIND closest to energy co2 emissions
-               (dimSums(mselect(vm_emiTeMkt, all_enty="co2"), dim=3) 
-               # deduce vm_emiTeMkt by CCU CO2 from vm_emiCdr 
-               # (which is only deduced from vm_emiAll, the total CO2 emissions, in REMIND)
-               - (1-p_share_CCS)*mselect(-vm_emiCdr, all_enty="co2"))*GtC_2_MtCO2,
+  if(getSets(vm_emiTeMkt)[[3]] == "emiTe"){
+    out <- mbind(out,
+              setNames(
```
pfuehrlich-pik, line 522: The indentation is a little off here. I'd suggest in RStudio `ctrl+a` then `ctrl+i` to auto-indent everything, then `lucode2::autoFormat('R/reportEmi.R')` to remove unnecessary whitespace introduced by `ctrl+i`. This kind of auto-formatting can make it hard to see the actual changes, so I'd suggest you do this after everything else is done, so right before merging, in a separate commit.

pfuehrlich-pik (changes requested):
Looks good, although I think we can reduce code duplication. In some places you have the same block of code in the if and the else branch with only one difference: mselect uses emiTe instead of all_enty. I'd suggest you try to isolate the part that is different, for example:
```
selection <- if(getSets(vm_emiTeMkt)[[3]] == "emiTe")
  mselect(vm_emiTeMkt, emiTe="co2")
else
  mselect(vm_emiTeMkt, all_enty="co2")
```
This way you can just use `selection` in the code block, no need to duplicate the whole block. Please ask me me on mattermost if you need some further help with that :)

R/reportEmi.R
```diff
@@ -1231,48 +1237,66 @@ reportEmi <- function(gdx, output=NULL, regionSubsetList=NULL,t=c(seq(2005,2060,
...
-                         "Emi|CH4|++|Outside ETS and ESR (Mt CH4/yr)"),
-               # N2O
-               setNames( dimSums(mselect(vm_emiAllMkt, all_enty="n2o",all_emiMkt="ETS"), dim=3) * MtN2_to_ktN2O,
-                         "Emi|N2O|++|ETS (kt N2O/yr)"),
-               setNames( dimSums(mselect(vm_emiAllMkt, all_enty="n2o",all_emiMkt="ES"), dim=3) * MtN2_to_ktN2O,
-                         "Emi|N2O|++|ESR (kt N2O/yr)"),
-               setNames( dimSums(mselect(vm_emiAllMkt, all_enty="n2o",all_emiMkt="other"), dim=3) * MtN2_to_ktN2O,
-                         "Emi|N2O|++|Outside ETS and ESR (kt N2O/yr)"))
-  
+            # CO2
```
pfuehrlich-pik, line 1279: Indentation

R/reportExtraction.R
```diff
@@ -65,8 +65,12 @@ reportExtraction <- function(gdx,regionSubsetList=NULL,t=c(seq(2005,2060,5),seq(
   if(!is.null(grades)) { grades     <- grades[,t,mappegrad[which(mappegrad != "peur.1")]] }
   costfu_ex  <- costfu_ex[,t,petyex]
   pebiolc_demandmag <- pebiolc_demandmag[,t,]
-  p_cint          <- collapseNames(mselect(p_cint,all_enty="co2"))
-  getSets(p_cint) <- gsub("all_enty1","all_enty",getSets(p_cint))
+  if(getSets(p_cint)[[3]] == "emiTe"){
+      p_cint          <- collapseNames(mselect(p_cint,emiTe="co2"))
+  } else {
+      p_cint          <- collapseNames(mselect(p_cint,all_enty="co2"))
+      getSets(p_cint) <- gsub("all_enty1","all_enty",getSets(p_cint))
```
pfuehrlich-pik, line 72: Is it correct that this is only in the `else` branch?

R/reportMOFEX.R
```diff
@@ -69,8 +69,12 @@ reportMOFEX <- function(gdx,gdx_ref=NULL,file=NULL,scenario='default') {
   fuel2pe.m <- fuel2pe.m[,t,]
   budget.m <- budget.m[,t,]
   tradebal.m <- tradebal.m[,t,]
-  p_cint          <- collapseNames(mselect(p_cint,all_enty="co2"))
-  getSets(p_cint) <- gsub("all_enty1","all_enty",getSets(p_cint))
+  if(getSets(p_cint)[[3]] == "emiTe"){
+      p_cint          <- collapseNames(mselect(p_cint,emiTe="co2"))
+  } else {
+      p_cint          <- collapseNames(mselect(p_cint,all_enty="co2"))
+      getSets(p_cint) <- gsub("all_enty1","all_enty",getSets(p_cint))
```
pfuehrlich-pik, line 76: Is it correct that this is only in the `else` branch?

pfuehrlich-pik (approved):
Please fix the indentation and consider my style suggestions. Otherwise, it's ready to merge I'd say :+1:

R/reportExtraction.R
spahrm: This correct since only in the else branch `"all_enty1"` needs to be renamed to `"all_enty"`, in the if branch no `"all_enty1"` exists anymore

R/reportMOFEX.R
spahrm: Yes see comment above

PR #103, opened 2021-09-03
fix climate plots in compareScenarios

giannou:


pfuehrlich-pik (approved):
Looks good. I think you are using an outdated lucode2, which is probably why buildLibrary removed the r-universe badhge in the README

giannou:
I restored it manually, hope that's ok

PR #109, opened 2021-09-09
 Refactor variables depending on emiTe (same like #100)

spahrm:
This is the same PR like #100. Where I applied the fixes from Felix and Michaja. But where the buildLibrary and the autoFormat is still missing. (I will do this after you approved this PR) 

After this PR is merged I will delete the other one.

Inside REMIND several variable names will be refactored. This variables will then be depending on emiTe instead of all_enty.
By checking via getSets all variables, later depending on emiTe instead of all_enty , use emiTe and as a default and for backward compability the all_enty fallback is still implemented.

Since this is my first PR at PIK and I'm also not an R expert, please check the code carefully and tell me if I could improve the code or if I missed something critical.


spahrm:
After approval:

- [ ] build library
- [ ] auto format changed files

pfuehrlich-pik (approved):

PR #112, opened 2021-09-16
update and re-enable the test

pfuehrlich-pik:
This PR re-enables the partially rewritten test. The suggests dependency on `stringi` was removed and a new up-to-date gdx file is downloaded for the test.


tests/testthat/test-convGDX2mif.R
```diff
@@ -20,79 +20,79 @@ check_eqs <- function(dt, eqs, scope = "all", sens = 1e-8) {
 
     dt[, diff := total - get(names(eqs)[LHS])]
     if (nrow(dt[abs(diff) > sens]) > 0) {
-      fail(sprintf("Check on data integrity failed for %s", names(eqs)[LHS]))
+      fail(paste("Check on data integrity failed for", names(eqs)[LHS]))
```
Loisel, line 23: interesting. Just out of curiosity, what is the problem with `sprintf`? I use it a lot...

Loisel (approved):
Looks plausible.

tests/testthat/test-convGDX2mif.R
pfuehrlich-pik: Not really a problem, but there's a little caveat, see "Warning" in the documentation: https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/sprintf
Also it's much much less commonly used compared to paste and paste0 in the pik-piam codebase and the codebase gets easier to read if we use the same function for this kind of thing everywhere.

PR #119, opened 2021-10-15
Refactor emi all changes

spahrm:
Inside REMIND several variable names will be refactored. This variables will then be depending on emiTe or emiAll instead of all_enty.
By checking via getSets all variables, later depending on emiTe or emiAll instead of all_enty. This handles a backward compability since it uses emiTe/emiAll as the default and all_enty case as its fallaback.

I will add the buildLibrary after this PR is approved.


LaviniaBaumstark (approved):
fine, if your tests and buildLibrary() work

PR #125, opened 2021-11-02
Add EDGE-T reporting stock and sale totals

Loisel:
The sublevel sums (over technologies or vehicle sizes) are added to the reporting.


fschreyer:
Thx for that!

PR #130, opened 2021-11-25
Add some SDP variables

rauner:
Add some SDP variables that can be calculated from the fulldata.gdx


R/calcSDPIndustry.R
```diff
@@ -0,0 +1,44 @@
+addIndustry_UE_from_FE <- function(output) {
+
```
LaviniaBaumstark, line 2: you should check here that the variables exist in you output and if not, run the needed reporting-scripts

R/convGDX2MIF.R
```diff
@@ -95,12 +95,24 @@ convGDX2MIF <- function(gdx,gdx_ref=NULL,file=NULL,scenario="default",t=c(seq(20
...
+  
+  # reporting of SDP variables
+  message("running reportSDPVariables...")
+  tmp <- try(reportSDPVariables(output))  # test whether reportSDPVariables works
+  if(class(tmp)!="try-error") {
+    if(!is.null(tmp)) output <- tmp
+  } else {
+    message("function reportSDPVariables does not work and is skipped")
+  }
+
```
LaviniaBaumstark, line 107: do I get it right: the output is not added but overwritten?

R/calcSDPTransport.R
```diff
@@ -0,0 +1,62 @@
+## calculating the UE from the FEs manually is not required anymore
+  ## once the UE variable in REMIND is reporting real UE -- AD 17-09-20
+
+addUEperCapForTransport <- function(output){
+
```
LaviniaBaumstark, line 5: here you also would need to check for variables existing

R/reportSDPVariables.R
```diff
@@ -0,0 +1,42 @@
...
+#' 
+#' @importFrom magclass setNames mbind mselect 
+#' @importFrom dplyr filter
+
+
+reportSDPVariables <- function(output=NULL) {
+  
+  
+  #from "calcSDPTransport.R"
+  output <- addUEperCapForTransport(output)
```
LaviniaBaumstark, line 25: why did you split all of the variables out into separate functions?

R/convGDX2MIF.R
LaviniaBaumstark: I would prefer calculating the new variables in your new function an adding it to the existing output instead of doing the adding in each single sub-function

PR #135, opened 2021-12-08
Useful energy reporting for buildings

rkrekeler:
If the GDX contains the UE demand in buildings, it will now be reported also for simple buildings.


fschreyer:
Hey Robin, seems good to me!

For my taste, you could comment your code a bit more and be explicit about what the variables mean as people who just might want to look up the meaning of the variables from the mif file and don't know this part of the mode well could go through this section for an explanation. Also, it might be worth mentioning that you use the fe2ue efficiencies from the baseline run to calculate those of the respective run at hand based on the FE output of the model.

rkrekeler:
> For my taste, you could comment your code a bit more and be explicit about what the variables mean as people who just might want to look up the meaning of the variables from the mif file and don't know this part of the mode well could go through this section for an explanation. Also, it might be worth mentioning that you use the fe2ue efficiencies from the baseline run to calculate those of the respective run at hand based on the FE output of the model.

Thanks Felix, well taken. I extended the commenting.

fschreyer (approved):

PR #136, opened 2021-12-13
Emi|GHG/CO2 subsector plots using CEDS UNFCCC data

pweigmann:
New compareScenario plots for 
- Emi|GHG|\<sector\>
- Emi|CO2|\<sector\>|\<subsector\>

... because CEDS2021 and UNFCCC now contains data to compare to.

Changed default: zeros are replaced by "NA" in historical data.

Increased compatibility with old _historical.mif_s by expanding conditional statements before plots.


R/compareScenarios.R
```diff
@@ -1603,33 +1601,79 @@ compareScenarios <- function(mif, hist,
   targets = c("Emi|GHGtot|target|40% (Mt CO2-equiv/yr)"="-40% by 2030 (vs. 1990)",
               "Emi|GHGtot|target|55% (Mt CO2-equiv/yr)"="-55% by 2030 (vs. 1990)",
               "Emi|GHGtot|target|65% (Mt CO2-equiv/yr)"="-65% by 2030 (vs. 1990)")
+  
+  # remove as soon as renaming is done
+  if ("Emi|GHGtot (Mt CO2-equiv/yr)" %in% getNames(histData, dim=3)) {
```
fschreyer, line 1606: Isn't this always true as some sources with ``Emi|GHGtot`` still hang around?

R/compareScenarios.R
```diff
@@ -1603,33 +1601,79 @@ compareScenarios <- function(mif, hist,
...
                          ylab='Total GHG Emisions w/ Bunkers [Mt CO2-equiv/yr]',scales="free_y",plot.priority=c("x_hist","x","x_proj"))
   swfigure(sw,print,p,sw_option="height=8,width=8")
-  
+
   p <- mipLineHistorical(data[,,"Emi|GHG|w/ Bunkers (Mt CO2eq/yr)"][mainReg,,,invert=TRUE],x_hist=NULL,
                          ylab='Total GHG Emisions w/ Bunkers [Mt CO2-equiv/yr]',scales="free_y",plot.priority=c("x_hist","x","x_proj"),facet.ncol=3)
   swfigure(sw,print,p,sw_option="height=9,width=8")
 
+  ## ---- Emissions GHG Sectors ----
+  items <- c("Emi|GHG|Energy (Mt CO2eq/yr)",
```
fschreyer, line 1635: Would be good to plot Emi|GHG|Land-Use Change here as well although there is no historic data for this yet to have all top-level IPCC sectors in this section.

R/compareScenarios.R
```diff
@@ -1698,228 +1742,302 @@ compareScenarios <- function(mif, hist,
...
   
-  swlatex(sw,"\\subsubsection{Industry}")
-  p <- mipLineHistorical(data[mainReg,,"Emi|CO2|Energy|Demand|Industry (Mt CO2/yr)"],x_hist=histData[mainReg,,"Emi|CO2|Industry|Direct (Mt CO2/yr)"],
-                         ylab='Emi|CO2|Energy|Demand|Industry [Mt CO2/yr]',scales="free_y",plot.priority=c("x_hist","x","x_proj"))
-  swfigure(sw,print,p,sw_option="height=8,width=8")
-  p <- mipLineHistorical(data[,,"Emi|CO2|Energy|Demand|Industry (Mt CO2/yr)"][mainReg,,,invert=TRUE],x_hist=histData[,,"Emi|CO2|Industry|Direct (Mt CO2/yr)"][mainReg,,,invert=TRUE],
-                         ylab='Emi|CO2|Energy|Demand|Industry [Mt CO2/yr]',scales="free_y",plot.priority=c("x_hist","x","x_proj"),facet.ncol=3)
-  swfigure(sw,print,p,sw_option="height=9,width=8")
+    swlatex(sw,"\\subsubsection{Energy Supply}")
+    p <- mipLineHistorical(data[mainReg,,"Emi|CO2|Energy|Supply (Mt CO2/yr)"],x_hist=histData[mainReg,,"Emi|CO2|Energy|Supply (Mt CO2/yr)"][,,"historical.CEDS.Emi|CO2|Energy|Supply (Mt CO2/yr)"],
```
fschreyer, line 1822: Why is the ``historical.CEDS`` necessary here? Wouldn't we want to plot from all sources which contain Emi|CO2|Energy|Supply?

fschreyer (changes requested):
Cool, looks good!

Was not entirely sure from taking a brief look. Are there plots doubled in the compare scenario PDF when CEDS/UNFCCC historical data is available, so once with and once without? That would actually be a bit confusing.

R/compareScenarios.R
pweigmann: EDGAR, UNFCCC and UBA also have data here, but not for all regions, resulting in reduced GLO numbers. This was my (not very elegant) solution to only plot those sources that have data for all regions.
image

R/compareScenarios.R
pweigmann: on it

R/compareScenarios.R
fbenke-pik: FYI UNFCCC problem with zeros has been fixed in mrremind on friday and should no longer occur. i will take care of UBA soon as well.

fbenke-pik: So I suggest fixing this in mrremind instead of here. Happy to provide an UBA fix asap.

pweigmann: Sounds good, in that case I will remove the filter for certain sources.

R/compareScenarios.R
pweigmann: True, I'll see if I can find a better solution.

pweigmann:
> Are there plots doubled in the compare scenario PDF when CEDS/UNFCCC historical data is available, so once with and once without?

I don't think there is. In case there are doubled plots it's a mistake from my side.

pweigmann:
I found a better way to include both Emi|GHGtot and Emi|GHG in the plots, added a Emi|GHG|Land-Use Change plot and removed the filters that took care of regionally incomplete data as Falk will fix that on the mrremind side. I didn't find any doubled plots either, so I think we might be good to go?

PR #137, opened 2021-12-13
change accounting of synfuel carbon in emissions to be accounted as emissions for the CO2 provider and adapt definition of gross emissions to include CO2 storage from synfuels

fschreyer:
This adapts the emissions reporting in several ways

* carbon used for CCU (synfuels) is now accounted as emissions for the CO2 provider (before it was accounted as emissions for the synfuel producer)
- this requires adding CCU carbon to pe2se carbon capture technologies and industry carbon capture which affects energy supply industry and industrial process emissions
* gross emissions are now defined as including carbon stored from captured synfuels in industry (in analogy to industry BECCS which includes carbon stored from biogenic carriers)
* correspondingly another CDR variable is added (to balance the gross emissions increase) (``Emi|CO2|CDR|Industry CCS|Synthetic Fuels (Mt CO2/yr)``)
* corresponding cumulative emissions and plots are added to compare scenario script
* Emi|CO2|Energy|Supply|Fuels variable is added for comparison to UNFCCC data

Overall, the change in synfuel carbon accounting and gross emissions now makes industry gross emissions always positive.


amerfort (approved):
Looks good to me.

PR #159, opened 2022-01-19
update compareScenarios2() to accept relative paths for mifScen, mifHist, and userSectionPath

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
So far, `knitr` is looking for files relative to wherever under `.libPaths()` it is installed.

(I will add the `lucode2::buildLibrary()` hubbub after approval to avoid additional conflicts.)


chroetz:
Thanks, this is very helpful!

PR #160, opened 2022-01-19
update compareScenarios2/cs2_main.Rmd to remove spurious warning

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
- if `mifScen` is unnamed, `mifScenNames` is `NULL`
- `tibble()` does not create `NULL` columns, so the column `newScenarioName` is dropped
  silently
- testing if `newScenarioName` is `NULL` generates
  > Warning: Unknown or uninitialised column: `newScenarioName`.
- testing for the existence of `newScenarioName` does not

(I will add the `lucode2::buildLibrary()` hubbub after approval to avoid additional conflicts.)

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '31748370'
+ValidationKey: '31938480'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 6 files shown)


chroetz:
Yes, that makes more sense!

PR #166, opened 2022-01-24
reporting for sector-differentiated CO2 taxes for NGFS project

orichters:
This is still a draft, some details still to be discussed.

In [remind PR 632](https://github.com/remindmodel/remind/pull/632), the option to use `p21_CO2TaxSectorMarkup` to implement sector-specific CO2 markups for buildings and transport was introduced. This PR integrates this option into the remind reporting. It calculates an average CO2 price per region, weighting the individual prices with either final energy or gross CO2 emissions, similar to the weighting used for determining the global CO2 price.

Additionally, the individual sector taxes are reported, which is required for the NGFS project.


R/reportPrices.R
```diff
@@ -316,8 +317,67 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
...
+  if (all(p21_CO2TaxSectorMarkup == 0)) { # then all are identical
+    out <- mbind(out, setNames(abs(pm_pvpRegi / (pm_pvp[,,"good"] + 1e-10)) * 1000 * 12/44, "Price|Carbon (US$2005/t CO2)"))
+    for (pcname in c("Price|Carbon|Demand|Buildings (US$2005/t CO2)", "Price|Carbon|Demand|Transport (US$2005/t CO2)",
+                     "Price|Carbon|Demand|Industry (US$2005/t CO2)", "Price|Carbon|Supply (US$2005/t CO2)",
+                     "Price|Carbon|AggregatedByGrossCO2 (US$2005/t CO2)")) {
+      out <- mbind(out, setNames(out[, , "Price|Carbon (US$2005/t CO2)"], pcname))
+    }
+  } else {  # currently p21_CO2TaxSectorMarkup is only implemented for build and trans in REMIND
+    out <- mbind(out,
+                 setNames((1 + p21_CO2TaxSectorMarkup[, , "build"]) * abs(pm_pvpRegi / (pm_pvp[,,"good"] + 1e-10)) * 1000 * 12/44,
```
fschreyer, line 335: I would be curious to know why ``abs(pm_pvpRegi / (pm_pvp[,,"good"])`` gives the carbon price in REMIND. 

``pm_pvpRegi`` is defined over all traded goods in the nash market, right?

Under which circumstances is it incorrect to take ``pm_taxCO2eq`` as the (regional) carbon price which is what is used in the [tax equation](https://github.com/remindmodel/remind/blob/d2917b5915430916f6f70992d4d1391923756226/modules/21_tax/on/equations.gms#L59)?

If it is always the same, it might be easier for people to understand if you take the parameter from the tax equation.

R/reportPrices.R
```diff
@@ -316,8 +317,67 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
       setNames(abs(esm2macro.m[,,name_trsp[1]]/(budget.m+1e-10)) * tdptwyr2dpgj , "Price|Energy Service|Transport LDV (US$2005/GJ)"))
   }
 
+  FE_data   <- reportFE(gdx, regionSubsetList = regionSubsetList, t = t)
+  Emi_data <- reportEmi(gdx, regionSubsetList = regionSubsetList, t = t)
+  FE_data_regi  <-  FE_data[getItems(pm_pvpRegi, dim = "all_regi"), , ] # get rid of GLO
+  Emi_data_regi <- Emi_data[getItems(pm_pvpRegi, dim = "all_regi"), , ]
+
+  # report GHG taxes, differentiated by sector
+  if (all(p21_CO2TaxSectorMarkup == 0)) { # then all are identical
```
fschreyer, line 326: Please also check that the new parameter ``p21_CO2TaxSectorMarkup`` exists with ``is.null(p21_CO2TaxSectorMarkup)``. We usually do this for some backwards compatability such that reporting does not get broken if you have not pulled REMIND for some days. However, maybe we can drop this requirement if ``p21_CO2TaxSectorMarkup`` really exists by default in all new runs, but I think that would be up to RSE to decide more generally here. I usually still implement this conditionally to not break other people's reporting.

fschreyer (changes requested):

R/reportPrices.R
christophbertram: pm_pvpRegi in reportPrices is only the "perm" component, and this gets set to the combined tax rate for tax runs: https://github.com/remindmodel/remind/blob/develop/core/postsolve.gms#L16
So the reason to not use only pm_taxCO2eq is that we by now have a couple of parameters, the sum of which is the effective carbon price.

R/reportPrices.R
orichters: If `p21_CO2TaxSectorMarkup` does not exist in the gdx file, then `readGDX` sets it to `NULL`. And `all(NULL == 0)` is `TRUE`, so your case is already covered.

christophbertram (approved):
Looks good to me then.

fschreyer (approved):

PR #177, opened 2022-01-27
fix fe81469 that caused "invalid subscript type 'closure'" errors

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
`t` is not defined, so still a function.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '33473440'
+ValidationKey: '33492459'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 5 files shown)


Renato-Rodrigues (changes requested):
I would add the t parameter to the main function, as it is done in other report functions like reportEmi:
```
reportSDPVariables <- function(gdx,output=NULL,t = c(seq(2005, 2060, 5), seq(2070, 2110, 10), 2130, 2150)) {
```
and restrict the output for the t dimension only.

Renato-Rodrigues (changes requested):
To use the output work around, I would say it is also necessary to add a conditional at the beginning of the reportSDPVariables to oblige that output is not null

```
if(is.null(output)){
    stop("please provide a file containing all needed information")
  }
```

If not, whenever you run directly the function and no output object is found, the getYears(output) will be empty by definition. As the function requires variables from reportFE and reportCrossVariables this would be a better error checking in the future.

PR #183, opened 2022-01-28
fix ab72bd4 where reportFE() would fail on incomplete .gdxes passed t…

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
…o exoGAINSAirpollutants.R


orichters (approved):
this should indeed fix this error. Thanks a lot for the debugging…

PR #188, opened 2022-02-09
Change Carbon Management variable names and minor bug fix

amerfort:
"Carbon Management|Carbon Sources" was renamed to "Carbon Management|Carbon Capture" 
"Carbon Management|Carbon Sinks" was split up to "Carbon Management|Storage" and "Carbon Management|Usage"
Enhanced Weathering is not part of the Carbon Management variable tree anymore, as this carbon does not enter the value chain.
Bugfix: "Emi|CO2|CDR|Industry CCS|Synthetic Fuels" now only accounts for the atmospheric share of the carbon contained in the synthetic fuels used in industry ccs applications.

### DIFF
--- a/R/reportEmi.R
+++ b/R/reportEmi.R
@@ -25,8 +25,7 @@
 
 
 reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(2005, 2060, 5), seq(2070, 2110, 10), 2130, 2150)) {
-
-
+  
 
   # emissions calculation requires information from other reporting functions
   if (is.null(output)) {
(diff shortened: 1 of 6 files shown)


fschreyer (commented):
Thanks. 

One comment/question: If we also change the gross variables and deduce the fossil synfuel industry CCS here, then Emi|CO2|Gross|Energy|Industry can become negative, in theory. Because fossil CO2 in synfuels is accounted as emissions of the CO2 provider (e.g. energy supply emissions) and not as industry emissions. So if industry does CCS with this carbon, then it is negative for industry. 

We could also leave the gross variables as they are and only change the CDR one.

fschreyer (approved):

R/reportEmi.R
```diff
@@ -1001,65 +996,54 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
...
   out <- mbind(out,
 
                # total captured and stored by pe2se BECCS
-               setNames(out[, , "Carbon Management|Carbon Sources|+|Biomass|Pe2Se (Mt CO2/yr)"] * p_share_CCS,
-                          "Carbon Management|Carbon Sinks|Storage|+|Biomass|Pe2Se (Mt CO2/yr)"),
-               setNames(out[, , "Carbon Management|Carbon Sources|+|Fossil|Pe2Se (Mt CO2/yr)"] * p_share_CCS,
-                          "Carbon Management|Carbon Sinks|Storage|+|Fossil|Pe2Se (Mt CO2/yr)"),
-               setNames(out[, , "Carbon Management|Carbon Sources|+|Industry Energy (Mt CO2/yr)"] * p_share_CCS,
-                          "Carbon Management|Carbon Sinks|Storage|+|Industry Energy (Mt CO2/yr)"),
+               setNames(out[, , "Carbon Management|Carbon Capture|+|Biomass|Pe2Se (Mt CO2/yr)"] * p_share_CCS,
```
strefler, line 1026: I think it would make sense to call the storage and usage variables "Carbon Management|Carbon Capture|Storage" and "Carbon Management|Carbon Capture|Usage". This would make it clear that "Carbon Management|Carbon Capture" = "Carbon Management|Carbon Capture|Storage" + "Carbon Management|Carbon Capture|Usage"

R/reportEmi.R
```diff
@@ -1153,45 +1136,53 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
...
+               setNames(-v33_emiEW * GtC_2_MtCO2,
                         "Emi|CO2|CDR|EW (Mt CO2/yr)"))
+  
+  out <- mbind(out,
+               # total CDR
+               setNames( out[, , "Emi|CO2|CDR|Land-Use Change (Mt CO2/yr)"] 
+                        + out[, , "Emi|CO2|CDR|BECCS (Mt CO2/yr)"] 
+                        + out[, , "Emi|CO2|CDR|DACCS (Mt CO2/yr)"]
+                        + out[, , "Emi|CO2|CDR|EW (Mt CO2/yr)"]                        
+                        + out[, , "Emi|CO2|CDR|Industry CCS|Synthetic Fuels (Mt CO2/yr)"],
```
strefler, line 1184: what about industry BECCS? Is that part of the "normal" BECCS? Maybe we should separate this out, too, if we have industry CCS|Synthetic Fuels.

PR #198, opened 2022-02-24
Add calculation of liquid shares in EDGE-T 

fbenke-pik:
Add split into Biomass/Fossil/Hydrogen for FE|..|Liquids variables according to FE demand shares. The implementation uses magclass instead of data.table, since this is the library used in most report functions. But the new code encapsulated in a function to create a separation between the two paradigms.


R/reportEDGETransport.R
```diff
@@ -288,15 +289,118 @@ reportEDGETransport <- function(output_folder=".",
...
+      datatable[!is.na(aggr_veh), sum(demand_EJ, na.rm = T),
+        by = c("region", "year", "aggr_veh", "remind_rep")
+      ][, variable := paste0(aggr_veh, "|", remind_rep)][,c("region", "year", "variable", "V1")]
+    )
+    
+    # splits Liquids variables into Biomass, Fossil and Hydrogen according to FE demand shares
+    .split_liquids <- function(df) {
+      demFeSector <- readGDX(gdx, "vm_demFeSector", field = "l", restore_zeros = F)
+
+      # biomass share in transport sector
```
fschreyer, line 306: I would write biomass share in biomass+hydrogen liquids in total transport sector

and same below... 

So, this only applies same shares of biomass within hydrogen+biomass per sector not within biomass+hydrogen+fossil. The latter can differ in runs with different carbon prices on different transport sectors.

fschreyer (approved):

R/reportEDGETransport.R
fbenke-pik: thanks, makes sense. i just updated the comments

MariannaR (approved):

Issue #195, opened 2022-02-24
Scenario Comparison 2 does not work with two runs with the same scenario name


Renato-Rodrigues:
How to reproduce the issue: run compareScenarios2, select any group of runs with duplicated names. See the plots in the pdf merging the run results from two different runs in the same plot line. 

Expected behavior. The "old" compareScenarios added a suffix to any duplicated scenario name to avoid the problem, ex: "DIAG-NetZero" and "DIAG-NetZero_2" results were plotted in independent lines.

Ex: I selected two different DIAG-NetZero scenarios but both scenarios are merged in the plot lines:
![image](https://user-images.githubusercontent.com/33692508/155530618-84b1b159-fbe8-44b9-8a2b-5aae3fb9a017.png)

chroetz:
As a workaround, you can rename scenarios in the cs2 function. If you have two scenarios both named "scen" with mifs located at "path1/to/scen.mif" and "path2/to/scen.mif" respectively, you can rename them to "scen_abc" and "scen_xyz", respectively, as follows:
```
compareScenarios2(
  mifScen = c("scen_abc" = "path1/to/scen.mif", "scen_xyz" = "path2/to/scen.mif"),
  mifHist = "path/to/historical.mif")
```
I wonder whether this actually is the better solution, as for automatically adding a suffix "_1" or "_2", you do not know which one is which. So explicitly renaming them should be preferred. 
Nonetheless, cs2 should at least give you a warning. And maybe, the warning message could state that you can do what I described above. 
@Renato-Rodrigues : Should cs2 abort if it detects duplicated scenario names to allow the user to directly specify the name or should it just add a default suffix like "_1" and "_2"?

Renato-Rodrigues:
Hi Christof, thks for the fast reply!

From my point of view, calling by hand compareScenarios defeats the purpose of having an automated Rscript output.R to speed up our process.
Also, renaming the scenarios explicitly is not a solution as it would require continuous renaming whenever we want to run a comparison with more up to date data or newer model versions.

Knowing which runs the suffix referred to was never a problem before for us as they always followed the same order you chose when calling the compareScenarios function. Besides, adding the suffix was the way it always worked, and clearly merging the data together is not the right way to handle that. I am also against just disconsidering the duplicated scenario names as this defeats the purpose of the scenarios that you want to compare. 

[...]

PR #199, opened 2022-02-25
cs2: duplicated scen name, choose sections by numbers, arg envir 

chroetz:
This PR essentially contains 4 changes to compareScenarios2:
1. If scenario names are the same in different mifs, a suffix "_1", "_2", ... is added to distinguish them.
2. A reference table is shown at the beginning of the document, which lists the file paths, the scenario names in the mifs, and the scenario names used in the document. See new section file cs2_00_info.Rmd.
3. A numeric vector is accepted as value of the yaml argument sections, e.g. `c(4, 2)` to show only the sections "04_abc" and "02_xyz" (in that order).
4. A new argument envir is added. It determines the environment which `rmarkdown::render()` uses to execute code chunks. This may be set to `globalenv()` with sections set to `NULL` so that cs2_main.Rmd loads an preprocesses data in the users global environment, without creating plots. This may simplify the processes of testing changes made to the cs2 Rmd files, as a separate copy of the Rmd files is not required anymore.

See also #195 for 1. and 2.


Renato-Rodrigues (approved):

inst/markdown/compareScenarios2/cs2_main.Rmd
```diff
@@ -70,40 +70,78 @@ if (interactive()) {
...
+}
+# Add suffixes "_1", "_2", ... in case of duplicated scenario names.
+# Side note: In the unlikly case, one has scenario names
+#   scenA, scenA, scenA_1, scenB,
+# they will be renamed to
+#   scenA_1, scenA_2, scenA_1, scenB
+# in the first iteration of the following loop. After the second iteration we get
+#   scenA_1_1, scenA_2, scenA_1_2, scenB.
+while (anyDuplicated(dataScenNested$newScenarioName)) {
+  warning("There are dupicated scenario names. They will be renamed.")
```
orichters, line 109: dupicated -> duplicated :)

Great you thought about this possibility with getting a duplicated name through the replacement process! An alternative solution would be to add `_1_` or `.1` to the scenario names, as scenario names never end with `_` and never contain a `.`, see https://github.com/remindmodel/remind/blob/develop/start.R#L262. But: minor stuff, all good!

orichters:
Don't have the knowledge for a full review, but thanks for your work, very appreciated!

inst/markdown/compareScenarios2/cs2_main.Rmd
chroetz: My code is not really a super clean solution and you made a good point, I haven't thought about these naming restrictions. Unfortunately, I allowed the user to set newScenarioNames without such restrictions. (The restrictions could be easily implemented, though.) So, a user could specify names ending in `_` or containing a `.`. 
But if a user is specifying such adversarial scenario names as in my example by hand, they must be evil anyway...
One could also use `base::make.unique()`, which would probably be the cleanest coding solution, but it transforms `c("a", "a")` to `c("a", "a_1")` instead of `c("a_1", "a_2")` which is a little less nice.
To not waste too much time on a problem that will probably never actually surface, I will just keep it how it is for now, I think.

orichters: I agree. The make.unique solution would have been ok for me as well. And as I said, great that you thought about this problem in the first place.

Issue #204, opened 2022-02-28
file reference table messes up with GHG emissions charts


Renato-Rodrigues:
The table is not fitting in the page for large paths, and it crops the first charts in the pdf.

You could add a page break to avoid the second issue.

Also, if possible could you double check the links of the table of contents and the charts? They are not matching right now.

![image](https://user-images.githubusercontent.com/33692508/156016997-529bda60-c00d-4184-ab5f-3b293cc088d6.png)

PR #206, opened 2022-03-01
cs2 print config 

chroetz:
`compareScenarios2()` gains a YAML arguments `cfgScen` and `cfgDefault` to optionally provide paths to the `config.Rdata` files of the scenarios and the `default.cfg`. Information from these files will be shown at the beginning of the output file. In particular, the `description` of the scenarios (see https://github.com/remindmodel/remind/pull/685) will be shown.

Furthermore, faulty links from the table of contents in the PDF output are fixed. This was mentioned here #204. To fix this, the order of `\usepackge{}` calls in the TeX code had to be changed: LaTeX-package `hyperref` needed to be included last. To achieve this, an explicit LaTeX template is added (a modified copy of [jgm/pandoc/data/templates/default.latex](https://github.com/jgm/pandoc/blob/581c94913fa055fa0d925db893991f01647b37f6/data/templates/default.latex)) for PDF of output which puts the inclusion of the `hyperref` at the end the header.


Renato-Rodrigues:
Very nice work!!!

By looking at the pdf example, is it possible to remove the unnecessary warnings?
```
## Warning: One or more parsing issues, see `problems()` for details
## One or more parsing issues, see `problems()` for details
## One or more parsing issues, see `problems()` for details
## One or more parsing issues, see `problems()` for details
```

And as a suggestion, I would only include at the beginning of the scenario config the sections `0.1 File Reference Table` and `0.2 Descriptions`, and move the scenario config analysis to the end of the file so we can jump more quickly from the reference table to the summary plots.

orichters:
Sorry, I can't provide a review, just saying that the results look as intended and thanks for your work…

chroetz:
> And as a suggestion, I would only include at the beginning of the scenario config the sections `0.1 File Reference Table` and `0.2 Descriptions`, and move the scenario config analysis to the end of the file so we can jump more quickly from the reference table to the summary plots.

With https://github.com/pik-piam/remind2/pull/206/commits/e948e63b7e01de1f96af9871bdb9bd26b1d32bfd, a "Further Infomation" section is put at the end of the output file. The config analysis is moved from the beginning to this new section.

Renato-Rodrigues (approved):

PR #214, opened 2022-03-10
for carbon tax revenue use net, not gross emission

orichters:
you can merge it directly


christophbertram (approved):
Perfect, thanks a lot

PR #215, opened 2022-03-11
cs2: automatic line break in long words with underscore, dash, ... 

chroetz:
Added `\usepackage{hyphenat}` to LaTeX inculdes to allow for line breaks in long words with underscore, dash, ... , in particular in the path column of the file reference table. See https://github.com/pik-piam/remind2/issues/204
Furthermore, `c_expname` and `c_description` are not shown in the Further Info section anymore as this is redundant.

### DIFF
--- a/inst/markdown/compareScenarios2/cs2_99_further_info.Rmd
+++ b/inst/markdown/compareScenarios2/cs2_99_further_info.Rmd
@@ -120,7 +120,11 @@ if (exists("cfgDefault") && !is.null(cfgDefault) &&
   cat(
     "Following GAMS settings are the same in all scenarios and not present in the default file:\n\n")
 
-  showInTables(cfgGmsWideSame[setdiff(names(cfgGmsWideSame), names(cfgDefault$gms))])
+  bind_cols(
+    Source = "Scenarios",
+    cfgGmsWideSame[setdiff(names(cfgGmsWideSame), names(cfgDefault$gms))]
+  ) %>%
+  showInTables()
 }
 ```
 
--- a/inst/markdown/compareScenarios2/cs2_99_further_info.Rmd
+++ b/inst/markdown/compareScenarios2/cs2_99_further_info.Rmd
@@ -75,6 +75,7 @@ if (exists("cfgGms") && !is.null(cfgGms)) {
     left_join(select(fileReference, fileid, newScenarioName), by = "fileid") %>%
     select(-path, -fileid) %>%
     rename(scenario = newScenarioName) %>%
+    filter(! name %in% c("c_expname" , "c_description")) %>% 
     pivot_wider(names_from = name, values_from = value) %>%
     mutate(across(-scenario, unlist)) ->
     cfgGmsWide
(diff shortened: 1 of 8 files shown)


chroetz:
Example: image
Note that LaTeX does necessarily make line breaks in arbitrary character strings. This change only adds the potential of line breaks at certain special characters.

inst/markdown/compareScenarios2/cs2_pdf_header_include.tex
```diff
@@ -9,4 +9,4 @@
    {\thesubparagraph\ }
    {0pt}
    {}
-   
\ No newline at end of file
+\usepackage{hyphenat} % allow for automatic line breaks in words with - and _
```
orichters, line 12: is this sufficient? As far as I know, the hyphenat package allows to control line break points more precisely, but does it automatically add line-breaks at these places without specifying them there? From the documentation:
```
The \bshyp command can be used to print a backslash and have following hyphenation.
The \fshyp command can be used to print a forward slash and have following hyphenation.
```
An alternative I use for my academic files, is this in the latex preamble:
```
\makeatletter
%\g@addto@macro{\UrlBreaks}{\UrlOrds}
\g@addto@macro\UrlBreaks{\do\*\do\~\do\'\do\"\do\a\do\b\do\c\do\d\do\e\do\f\do\g\do\h\do\i\do\j\do\k\do%
\l\do\m\do\n\do\o\do\p\do\q\do\r\do\s\do\t\do\u\do\v\do\w\do\x\do\y\do\z\do\&\do\1\do\2\do\3\do\4\do\5\do\6\do\7\do\8\do\9\do\0\do\.}
\makeatother
```
This simply allows breaking in urls after any character. I particularly do not like breaking after `-` because then it is often unclear whether this is added as a line-wrap hyphen or whether is was part of the link.

inst/markdown/compareScenarios2/cs2_99_further_info.Rmd
```diff
@@ -75,6 +75,7 @@ if (exists("cfgGms") && !is.null(cfgGms)) {
     left_join(select(fileReference, fileid, newScenarioName), by = "fileid") %>%
     select(-path, -fileid) %>%
     rename(scenario = newScenarioName) %>%
+    filter(! name %in% c("c_expname" , "c_description")) %>%
```
orichters, line 78: great!

orichters (approved):
added some comments…

inst/markdown/compareScenarios2/cs2_pdf_header_include.tex
chroetz: Thanks for the suggestion. 

Just adding the package seems to work in the example and it is a very simple solution.

True, it is not perfect. But an improvement requires more complex code. For reference, I comment here, what I think one would need to do, but leave the code as it is for now.

* Define a macro as in @orichters comment above or use `\seqsplit{}` from the package `seqsplit`.
* Check whether the output is HTML or PDF via `getOption("knitr.table.format")`.
* For PDF output modify the respective data frame values by replacing their `"content"` by `"\seqsplit{content}"`.
* call `kbl()` with argument `escape = FALSE`

orichters: As always: If you checked it and it works, it is fine. I am just surprised it does, that's all. So leave it as it is.

PR #218, opened 2022-03-21
PFC emissions

pweigmann:
Add PFC variables

Emi|CF4 (kt CF4/yr)
Emi|C2F6 (kt C2F6/yr)
Emi|C6F14 (kt C6F14/yr)

to reporting.


christophbertram (approved):
Perfect

PR #219, opened 2022-03-22
Add FE|w/o Non-energy Use|Industry to reportFE

fbenke-pik:
Report FE|w/o Non-energy Use|Industry with Non-energy Use as a temporary fix to ensure inclusion in SHAPE submission.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '34958976'
+ValidationKey: '34979882'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.83.3",
+  "version": "1.83.4",
   "description": "<p>Contains the REMIND-specific routines for data and model output manipulation.<\/p>",
   "creators": [
     {
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: remind2
 Type: Package
 Title: The REMIND R package (2nd generation)
-Version: 1.83.3
-Date: 2022-03-21
+Version: 1.83.4
+Date: 2022-03-22
 Authors@R: as.person(c(
     "Renato Rodrigues <renato.rodrigues@pik-potsdam.de> [aut,cre]"))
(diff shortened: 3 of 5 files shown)


Renato-Rodrigues (approved):

Renato-Rodrigues (changes requested):
Just for completeness, could you add `FE|Non-energy Use|Industry` equal to zero also in this case?
Also, could you repeat the same code in an else under the condition inside:
```
if ("DEU" %in% getRegions(vm_prodFe)) {
 ... 
} else { 
same code here to define `FE|w/o Non-energy Use|Industry (EJ/yr)` and `FE|Non-energy Use|Industry`.
}
```

fbenke-pik:
@Renato-Rodrigues 
(1) will do
(2) not necessary in my current adjustment, I merged the 2 conditions into one, since there were no actions if only the first condition is met

fbenke-pik:
@Renato-Rodrigues @silviamade  introducing "FE|w/o Non-energy Use|Industry" with all 0s causes new problems, because reportEmi uses the presence of this variable as condition that triggers additional calculations that cannot be done as of now, because the subvariables of FE|w/o Non-energy Use|Industry are not present. if you think we should look further into this, let's discuss and work on a new PR for the changes. merging this for now

Renato-Rodrigues:
@falk, thks! I also agree with merging the way it is now.
@silviamade Changing the mapping wouldn't solve either issue. If something is nonexistent in the remind reporting, the change needs to be made in the remind reporting. This affects the same way `FE|w/o Non-energy Use|Industry|Solids`, as it is not in the reporting yet as far as I know, and changing the mapping just creates an error that will be propagated to the future after these variables are correctly included to the reporting.

silviamade:
We do have the variables that add up correctly in the current reporting, i.e. FE||Industry|Solids, Gases etc. so if we map the correct variables everything adds up correctly. Introducing this new naming in the mapping -> FE|w/o Non-energy Use|Industry|Solids as someone did creates the discrepancies in the variables summation checks. So what I am saying it that right now I have to modify the mapping in any case, even if Falk revises the reporting and changes the variable name to FE|w/o Non-energy Use|Industry. Unless we revise the reporting for industry entirely and change all the sub-variables names from FE||Industry|Solids, Gases, Liquids to FE|w/o Non-energy Use|Industry|Solids, Gases, Liquids. I don't think this is trivial but I might be wrong

Renato-Rodrigues:
`FE||Industry|Solids` in the reporting include non energy use, no? and in the navigate template they shouldn't, or I got something wrong?

PR #221, opened 2022-03-24
Kaya ref year

chroetz:
Adds `yearRef` as YAML parameter with default `2020` to `compareScenarios2()`. It is used to show values in the Kaya decomposition relative to  a reference year. Before, this year was hard coded as `2005`, which would cause the respective code for the Kaya plot to show an error message instead of the plot. See also #8.

Now the smallest available year greater or equal to `yearRef` is used as reference year in the relative Kaya plots.

Examples:

* If neither `yearsScen` (the years shown for scenario data) nor `yearRef` is specified, nothing changes, i.e. `2020` is the reference.
* If `yearsScen = seq(2025, 2055, 5)`, then `2025` is automatically chosen as reference year.
* If `yearRef=2005`, then `2005` is used as reference year.
* If `yearsScen = seq(2010, 2055, 5)` and `yearRef = 2013`, then `2015` is chosen as reference year.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35098000'
+ValidationKey: '35290600'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.84.0",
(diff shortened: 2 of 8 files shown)


Renato-Rodrigues (approved):
I have limited knowledge about the Kaya decomposition to say if it even makes sense to use other years as reference. Alternatively you could also simply not include these charts in case the ref year is missing.
Maybe you could ask somebody from the macro group if they agree with changing the ref year like you did.
Besides that, from a coding point of view I see no problems in merging this pull request.

chroetz:
After talking to @nicobauer, the default value for the reference year was set to 2020.

PR #222, opened 2022-03-25
Add area plots for detailed FE prices

fbenke-pik:


inst/markdown/compareScenarios2/cs2_02_macro.Rmd
```diff
@@ -115,6 +115,104 @@ vars %>%
...
+  items <- c(
+    paste0(g, "|Fuel Cost"),
+    paste0(g, "|Transport and Distribution"), 
+    paste0(g, "|Carbon Price Component"), 
+    paste0(g, "|Other Taxes") 
+  )
+  
+  tot <- paste0(g, "|Total LCOE")
+  
+  showAreaAndBarPlots(dPrice, intersect(items, unique(dPrice$variable)), tot)
```
chroetz, line 144: It is not necessary to do a "safety-intersect" like that, as `showAreaAndBarPlots()` essentially calls `filter(dPrice, filter(variable %in% items))` and will ignore missing variables for the plot. But it will print a warning message telling you which variables are missing.

chroetz (approved):

inst/markdown/compareScenarios2/cs2_02_macro.Rmd
fbenke-pik: I actually removed it, because while it was not crashing, the variables not used were still showing up in the legend, which i found confusing.

chroetz: oh, ok, then that is a bug in `showAreaAndBarPlots()`

PR #225, opened 2022-04-04
Fix EW sign

chroetz:
`"Emi|CO2|CDR|EW (Mt CO2/yr)"` now has the correct sign


amerfort (approved):
Thanks for spotting and fixing it!

PR #232, opened 2022-04-20
add compareScenConf, rework colorScenConf

orichters:
this PR
- adds a function `compareScenConf` which can be used to compare two versions of a scenario config file, printing the differences. Such a function is helpful because [such a github diff is completely useless](https://github.com/remindmodel/remind/pull/758/files#diff-adad3fc7eaac664e058c1080b15caa133710e2f4cfae631310f2e188b707c244). Instead, one can add [such a text](https://github.com/remindmodel/remind/pull/758#issuecomment-1102888119) to a PR, which is much more human-readable. See also [pik-piam/discussions/8](https://github.com/pik-piam/discussions/discussions/8).
- adds some new features from REMIND to colorScenConf to avoid false positives
- moves the `get_line()` function to `utils.R`, an idea taken from [the mip package](https://github.com/pik-piam/mip/blob/master/R/utils.R)


R/compareScenConf.R
```diff
@@ -0,0 +1,110 @@
...
+#'
+#' @return nothing. Prints a lot of messages.
+#' @importFrom utils read.csv2
+#' @export
+compareScenConf <- function(fileList = "", configfile = "default.cfg", renamedColumns = NULL, renamedRows = NULL) {
+  if (!file.exists(configfile)) {
+    if (file.exists(paste0("config/", configfile))) {
+      configfile <- paste0("config/", configfile)
+    } else {
+      stop(paste0("No 'default.cfg' in working directory (", getwd(), ") or unknown 'configfile' in function call."))
```
mikapfl, line 26: An error message `"Config file '", configfile, "' not found in current directory or in the 'config/' subdirectory."` or something along those lines would be more useful, I think.

mikapfl (approved):
Looks pretty useful to me!

Ideally, you could also add a simple test to `tests/testthat/`, just comparing two CSVs and asserting that the output matches a certain format. Doesn't look too useful now, but will be very useful when we have to change something in five years and then we have an easy way to tell whether something broke or not.

R/compareScenConf.R
mikapfl: Or better yet, mention `getwd()` and `configfile` in the error message.

orichters: You are right, I changed it such that
```
compareScenConf(configfile = "quak")
```
now returns:
```
Error in compareScenConf(configfile = "quak") : 
  No configfile quak in working directory (/p/tmp/oliverr).
```

orichters:
Regarding the tests: As the function does not return anything, will not be called by any other function, and is solely designed to provide human-readable messages, I don't see how I should design a test for that.

mikapfl:
> will not be called by any other function
Why do you think we won't build a github action which will automatically render these diffs for pull requests in REMIND? Would certainly be useful.

orichters:
Ok, then I have to admit I have no idea how I would design such a test.

mikapfl:
Ok, I guess then we leave it as a problem for future-Mika or future-Oliver. Should it become necessary, my strategy would be to first enable a mode (maybe via an argument to the function) where the text is returned instead of generating `warning`s (could be non-default) and then write a traditional test using that mode. For my envisioned use case, a mode which returns a text instead of printing warnings would anyway be necessary, I think.

orichters:
Ok, it can't be that difficult. So I added a switch `printit` and also return a list of warning message abbreviations that can be used to analyze the output in a test functions, after writing some config files and small csv files to be analyzed.
```
ℹ Testing remind2
✔ | F W S  OK | Context
✔ |        13 | compareScenConf [0.2s]
```
Looks good.

mikapfl:
Great, thanks!

orichters:
Improved the searching of default.cfg a bit, so it also looks in the folder the user provided for searching the csv files, did some cleanup. Also it either prints or returns the output now, but always returns the warnings. If @Renato-Rodrigues agrees, I'm ready to merge.

PR #234, opened 2022-04-22
suppress 'Joining, by' messages that clutter the logs

orichters:
The `join` commands regularly clutter the logs like that:
```
[1] "start generation of LCOE reporting"
Joining, by = "tech"
Joining, by = c("region", "tech", "rlf")
Joining, by = c("region", "tech")
Joining, by = c("region", "fuel")
Joining, by = c("region", "tech")
Joining, by = "tech"
Joining, by = "region"
Joining, by = "gridtech"
Joining, by = c("region", "tech")
Joining, by = "tech"
Joining, by = "teStor"
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "tech")
Joining, by = c("region", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = "tech"
Joining, by = c("region", "period", "fuel")
Joining, by = c("region", "period")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "tech")
Joining, by = c("region", "tech")
Joining, by = c("region", "period")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period")
Joining, by = c("region", "period", "tech", "fuel")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "tech")
Joining, by = c("region", "period", "output")
Joining, by = c("region", "period", "tech", "sector")
[1] "end generation of LCOE reporting"
```
I added utility functions that suppress this message and used them everywhere, avoiding these not so helpful messages in the REMIND logs. Also, @chroetz, I added a warning message to `compareScenarios` that it is outdated.

### DIFF
--- a/R/compareScenarios.R
+++ b/R/compareScenarios.R
@@ -36,6 +36,8 @@ compareScenarios <- function(mif, hist,
                              y_bar=c(2010,2030,2050,2100),
                              reg=NULL, mainReg="GLO", fileName="CompareScenarios.pdf",
                              sr15marker_RCP=NULL) {
+  message("compareScenarios is outdated and may break - use compareScenarios2 instead!")
+
   lineplots_perCap <- function(data, vars, percap_factor, ylabstr,
                                global=FALSE, mainReg_plot=mainReg, per_gdp=FALSE, histdata_plot=NULL){
 
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35495232'
+ValidationKey: '35514336'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.85.8",
+  "version": "1.85.9",
   "description": "<p>Contains the REMIND-specific routines for data and model output manipulation.<\/p>",
   "creators": [
     {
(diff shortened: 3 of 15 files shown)


R/compareScenarios.R
```diff
@@ -36,6 +36,8 @@ compareScenarios <- function(mif, hist,
                              y_bar=c(2010,2030,2050,2100),
                              reg=NULL, mainReg="GLO", fileName="CompareScenarios.pdf",
                              sr15marker_RCP=NULL) {
+  message("compareScenarios is outdated and may break - use compareScenarios2 instead!")
```
orichters, line 39: @chroetz: here I added the warning message

chroetz: Thanks! That is helpful. 
I think, the function should be removed eventually.
Maybe it can even be removed now as it doesn't work on current mifs?

chroetz:
Thanks! Removing uninformative logs is very helpful.

But I'm not sure whether `_join_silent()`-functions are the way to go here. 
I think the intention of the authors of the `_join()`-functions is that you are supposed to always specify the `by` argument in package code. This makes it explicit on what variables the join happens and does not create messages. Using the function without by and with automatic detection of fitting columns is for interactive use of the function. 
It would be ok to write a specific join function like `join_LCOE_components()` that always joins on specific columns. But `suppressMessages()` (or even `suppressWarnings()`) should be used with care and probably not as a default.

orichters:
Thanks, Christof, you are probably right. I wouldn't suppress warnings, but I thought messages were not such a problem…

PR #235, opened 2022-04-22
compareScenConf more flexible, + deprecation message compareScenarios

orichters:
- make compareScenConf more flexible such that it can be used for [project_interfaces as well to generate such messages for a  PR](https://github.com/pik-piam/project_interfaces/pull/18), because there the problem of unreadable diffs is [even worse](https://github.com/pik-piam/project_interfaces/pull/6/files)
- add warning message to compareScenarios

### DIFF
--- a/R/compareScenarios.R
+++ b/R/compareScenarios.R
@@ -36,6 +36,8 @@ compareScenarios <- function(mif, hist,
                              y_bar=c(2010,2030,2050,2100),
                              reg=NULL, mainReg="GLO", fileName="CompareScenarios.pdf",
                              sr15marker_RCP=NULL) {
+  .Deprecated("compareScenarios2")
+
   lineplots_perCap <- function(data, vars, percap_factor, ylabstr,
                                global=FALSE, mainReg_plot=mainReg, per_gdp=FALSE, histdata_plot=NULL){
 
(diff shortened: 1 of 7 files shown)


R/compareScenarios.R
```diff
@@ -36,6 +36,8 @@ compareScenarios <- function(mif, hist,
                              y_bar=c(2010,2030,2050,2100),
                              reg=NULL, mainReg="GLO", fileName="CompareScenarios.pdf",
                              sr15marker_RCP=NULL) {
+  message("compareScenarios is outdated and may break - use compareScenarios2 instead!")
+
```
mikapfl, line 40: The standard way to deprecate a function is to use
`.Deprecated("compareScenarios2")`
That will generate the correct messages / warnings.

Is it useful to keep `compareScenarios`? Or would it be less confusing to remove it outright?

mikapfl (changes requested):
Looks good to me, thanks for caring about properly retiring old functions as well!

I just have a small suggestion about how to do it.

mikapfl (approved):

R/compareScenarios.R
orichters: I changed that to `.Deprecated`. If somebody wants to remove it, that is ok, but I want just avoid that people use it and get confused as today in the REMIND mattermost channel.

PR #237, opened 2022-04-26
add HFC, SF6 reporting

pweigmann:
two new variables added to reporting, which were already present in REMIND:
- Emi|HFC (kt HFC134a-equiv/yr)
- Emi|SF6 (kt SF6/yr)


orichters (approved):
I loaded a gdx and ran the lines 1490–1495 in reportEmi.R which worked flawlessly. The rest seems to be only deleting blanks. Thanks!

PR #240, opened 2022-04-27
update windoff grid factor in LCOE calculation

cchrisgong:


Renato-Rodrigues:
Always add the author of the code related with your changes as one of the reviewers when you make any pull request.
I did not work with LCOE grid costs representation.

One related question although, is this number reflecting some input data in the model? 
It is usually not good practice to hard code numbers in the reporting functions. If the data is available in any way under an specific variable or parameter in the remind code, you should instead read it from the gdx file instead of hard coding the value to avoid future incompatibilities.

cchrisgong:
Hi Renato, Felix is on parental leave, so I don't have an alternative. I need this for the paper I am writing.

The parameter is hard-coded in the model. Not read in as input data. I think we want to keep this in the code for now, since it is ad-hoc and is still being improved

Renato-Rodrigues:
You should always include the person responsible for the part of the code that you are changing, no matter if the person is capable of reviewing your PR or not. 
This was presented and discussed in the remind meeting. This is crucial to avoid surprising the author of the code with changes that would affect the original intent of the code.

Waiting for the proper review process does not limit you from working with your changes. You can create your own library versions with your changes and use them as you like.

I definitely cannot approve a change that I don't know what it does. I don't have the time to dig into all possible code details of remind and review all PRs when people is not available by whatever reason.

cchrisgong:
Thanks for the clarification. This is a very small change. I will tag Felix, but it will not impact other people's work as far as I can see

PR #241, opened 2022-04-28
fix reportTax for EU21

orichters:
also the data for `EUR`, `NEU` and `EU27` has to be removed from `output_wo_GLO` to avoid this error:
```
Error in mbind(out, setNames(output_wo_GLO[, , "Price|Carbon|Demand|Buildings (US$2005/t CO2)"] *  : 
  Cannot handle objects! Data as well as spatial dimensions differ!
```
Maybe we should add a `convGDX2MIF` for a EU21 file to the tests of remind2.

I ran `convGDX2MIF` with a file provided by pweigmann and it worked without flaws. Also the obligatory test with a H12 file went through.


Renato-Rodrigues (approved):

Renato-Rodrigues:
before merging you should be sure the code checks pass.

PR #244, opened 2022-05-02
fix: output contains no ++ here

orichters:
line 52 in reportPrices states: `output <- deletePlus(output)`
Therefore, the `+` and `++` later must be deleted. This bug has only impacted the NGFS runs with sector-differentiated carbon prices.
I tested the `d_rap` run and it works.


christophbertram (approved):
Looks solid, and seems like this is the default way to do it...

Issue #245, opened 2022-05-03
reportEDGETransport fails if _adjustedpolicycost.mif is present in folder


orichters:
If a file such as `REMIND_generic_SSP2-Base_adjustedPolicyCosts.mif` is present in the output folder, [these lines](https://github.com/pik-piam/remind2/blob/master/R/reportEDGETransport.R#L70:L74) don't select a single file and `readMIF` throws an error. The `stopifnot` check doesn't catch it because `typeof(c("1", "c"))` is also `character`.
```
[1] "start generation of EDGE-T reporting"
Error in fread(mif, header = T) : 
  input= must be a single character string containing a file name, a system command containing at least one space, a URL starting 'http[s]://', 'ftp[s]://' or 'file://', or, the input data itself containing at least one \n or \r
Calls: reportEDGETransport -> readMIF -> fread
Execution halted
```
Tagging the people that last worked on it. Thank you!

PR #246, opened 2022-05-03
Fix check for existing reporting file to append the EDGE-T reporting

Loisel:
(hopefully) addresses issue #245

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35628496'
+ValidationKey: '35649475'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35649475'
+ValidationKey: '35668590'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.86.4",
+  "version": "1.86.5",
   "description": "<p>Contains the REMIND-specific routines for data and model output manipulation.<\/p>",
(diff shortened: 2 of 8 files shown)


orichters (approved):
Thanks, @loisel, I tested it with both existing adjustPolicyCost runs and without and it solves the issue!

Did you delete the `miffile <- readMIF(name_mif)` on purpose? It seems not to be used later anyway.

On the other hand, if you just add further .mif files for whatever reasons into your output folder, it may break again. An alternative I think would be to add the option `file = NULL` to the function definition, do the search only if no such file is provided, and add `file = remind_reporting_file` [here](https://github.com/remindmodel/remind/blob/develop/scripts/output/single/reporting.R#L63).

But anyway, it fixes the problem that I had, and thanks a lot for your quick reaction!

Loisel:
Yes, that is a good idea! I thought about passing a parameter but decided to not break the interface. But having it optional would certainly do the trick! I consider returning the data (instead of appending) even more elegant. Maybe I would add a parameter `append=TRUE` which will search and append to mif for backwards compatibility but in the REMIND reporting script append the data explicitly.

Loisel:
> Did you delete the `miffile <- readMIF(name_mif)` on purpose? It seems not to be used later anyway.

Indeed, this line was an artifact.

orichters:
Looks good. So you plan to change [reporting.R](https://github.com/remindmodel/remind/blob/develop/scripts/output/single/reporting.R) in REMIND as well? Although I see it is not strictly necessary…

orichters (approved):

Loisel:
> Looks good. So you plan to change [reporting.R](https://github.com/remindmodel/remind/blob/develop/scripts/output/single/reporting.R) in REMIND as well? Although I see it is not strictly necessary…

Yes, I'll provide a PR as soon as the change hits master. The newer version of the EDGE-T reporting will return a quitte object, so this change in a sense is an acceptance test.

MariannaR (approved):

Loisel:
> Looks good. So you plan to change [reporting.R](https://github.com/remindmodel/remind/blob/develop/scripts/output/single/reporting.R) in REMIND as well? Although I see it is not strictly necessary…

Yes, I'll provide a PR as soon as the change hits master. The newer version of the EDGE-T reporting will return a quitte object, so this change in a sense is an acceptance test.

Issue #238, opened 2022-04-26
adapt test-convGDX2mif.R to compareScenarios2


orichters:
Is it really a good idea that `[test-convGDX2mif.R](https://github.com/pik-piam/remind2/blob/master/tests/testthat/test-convGDX2mif.R#L94) checks the mifs using the deprecated `compareScenarios`? Wouldn't it be preferable to use a more recent `fulldata.gdx` and check with `compareScenarios2`?

orichters:
Ok, thanks for the explanation. The fact is that `compareScenarios` is not working on current gdx files (which certainly is not good), which makes `compareScenarios2` the de facto standard for everyone using fairly recent REMIND runs, and it looked a bit strange to me that cs1 is used in the tests.

LaviniaBaumstark:
I would say: we can delete compareScenarios and to the test for compareScenarios2. this is the new standard and I think there is nothing missing which was provided by compareScenarios

chroetz:
The old `compareScenarios()` has an optional argument `sr15marker_RCP`, which is used in only one [plot](https://github.com/chroetz/remind2/blob/be6546cbb0ea9232442b250b8bc14df357c6781e/R/compareScenarios.R#L540). As far as I know, this is the only feature that is not implemented in `compareScenarios2()`.
The naming and order of arguments in cs2 is different from old cs. Furthermore, cs2 has additional arguments and features. 
I also favor just deleting the old cs and using cs2 in the test, even though having functions named `cs1`, `cs2`, `cs3`, ... is not the cleanest solution...

orichters:
my impression is that compareScenarios2 takes substantially longer that the old compareScenario. I would rather prefer not to wait even longer for `lucode2::buildLibrary`, it is already annoyingly long.

chroetz:
I agree, the test takes annoyingly long!

cs2 takes roughly the same time as cs1 to create the PDF output, at least this was the case when it was new. Creating HTML takes longer.

It's probably not necessary to test cs2 on all plots. E.g., we could just test the creation of the summary section or even just the info section at the beginning. In contrast to cs1, cs2 is tolerates errors and missing variables in the plots without failing completely. The plotting sections basically cannot fail. So, there is no gain in testing these.

My suggestion would be:

```
tmpDir <- tempdir()
compareScenarios2(
  mifScen = myMifs,
  mifHist = histMif,
  outputFormat = "pdf",
  outputFile = "cs2_test",
  outputDir = tmpDir,
[...]

PR #249, opened 2022-05-04
use new `compareScenarios2()` in test instead of old `compareSceanrios()`

chroetz:
See #238.

### DIFF
--- a/tests/testthat/test-convGDX2mif.R
+++ b/tests/testthat/test-convGDX2mif.R
@@ -90,9 +90,15 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
   if (!file.exists(histMif)) {
     utils::download.file("https://rse.pik-potsdam.de/data/example/historical.mif", histMif, quiet = TRUE)
   }
-  scenarioComparisonPath <- file.path(tempdir(), "scenarioComparison.pdf")
-  suppressWarnings(compareScenarios(myMifs, histMif, fileName = scenarioComparisonPath))
-  expect_true(file.exists(scenarioComparisonPath))
+  capture.output( # Do not show stdout text.
+    compareScenarios2(
+      mifScen = myMifs,
+      mifHist = histMif,
+      outputFormat = "pdf",
+      outputFile = "cs2_test",
+      outputDir = tempdir(),
+      sections = 0)) # Render only the info section.
+  expect_true(file.exists(file.path(tempdir(), "cs2_test.pdf")))
   unlink(tempdir(), recursive = TRUE)
   tempdir(TRUE)
 })


chroetz:
On my machine, the test now takes 3min.

orichters:
Thanks, Christof! May it be worth adding a second fulldata.gdx, maybe a H21 run (if that is possible in a cs2 to have both simultaneously?), or with sector-differentiated carbon prices (would have catched [this one](https://github.com/pik-piam/remind2/pull/244)), from a coupled run or something like that? So to say, design a second mif which is maximally different from the current one?

Renato-Rodrigues:
Could the library tests for alternative gdxs  be done in parallel? Maybe @giannou already know about that?
If not I am against adding additional gdx files to be tested for now. 
Current tests already take too long, so I would first focus on improving test performance/testing single regions instead, and so on, before making the accumulated number of tests take a very long time to finish.

orichters:
@Renato-Rodrigues: As far as I understand, @chroetz managed to bring the test time down to 3 minutes (convGDX2mif + cs2), so even if we add a second gdx, it would still be a big improvement.

Renato-Rodrigues:
The problem is keep summing up tests above tests without worrying about the total buildLibrary time.

If tests can run in parallel, I completely agree with adding as many tests as we can. 

I am not ok with having a lot of tests added sequentially because we need then to wait half hour or more for the buildLibrary to finsih to be able to commit an urgent model breaking change.
Maybe I am biased by the fact that when things get to me they are usually too urgent and need to be fixed asap, but this framework works very nicely for non urgent cases, but it is unbearable for cases like the one I am describing here.

orichters:
Yes, I agree. So: If we don't provide a second gdx file, but just a second mif file (that then has to be updated regulary to keep track of the changes in the reporting), this shouldn't extend the runtime of cs2 at all.

giannou:
We used to have parallelization in the tests, we can maybe reintroduce it. I think it was removed for reasons that are not relevant any more.

Renato-Rodrigues:
gdx files are not updated as frequently as mif files. We add or change reporting variables more often than model variables. So I don't think the mif file is a good approach.
I think the solution is to make sure that tests can run in parallel for multiple gdxs and scenario comparison tests. I am not even sure if this is the case already as I never checked the tests code.

I think we are going to far from the topic of the PR, maybe we should discuss that in an issue instead.

chroetz:
1) I timed the code again in more detail: I ran the test three times and it was always roughly 120s (Don't know why it was 180s before...)
* 85s `convGDX2MIF()`
* 25s `checkIntegrity()`
* 7s `compareScenarios2()`
* 3s other stuff

2) Should we parallelize using `parallel::mclapply()`?

3) Currently the test downloads <https://rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_fulldata.gdx> and <https://rse.pik-potsdam.de/data/example/historical.mif>. Are these files maintained / updated? How can additional gdxs be added?

orichters:
I suggest to merge that now because it is already a big relief, and discuss further which gdx files could be added, if this was desired.

orichters (approved):
I took the test and ran it without deleting the file and had a look at it. Seems to work and is much faster than before.

LaviniaBaumstark (approved):

PR #252, opened 2022-05-06
improve rownames in compareScenConf

orichters:
if `row.names = NULL` is passed, which is the option used for mapping files from `project_interfaces` that do not have a unique column identifier, put together something unique out of the first two columns instead of just numbering them.

Without that, if you add a single row, the whole matching procedure fails and all rows after the added row are considered as changed. With this PR, it looks like that:
```
~ 16: Carbon Sequestration|CCS|Biomass|Energy|Demand|Industry:
    r30m44: Emi|CO2|CDR|BECCS|Industry -> Carbon Management|Storage|Industry Energy|+|Biomass
~ 17: Carbon Sequestration|CCS|Biomass|Energy|Supply:
    r30m44: Emi|CO2|CDR|BECCS|Pe2Se -> Carbon Management|Storage|+|Biomass|Pe2Se
```

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35731542'
+ValidationKey: '355785980'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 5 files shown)


mikapfl (approved):
Nice addition!

PR #255, opened 2022-05-10
add CCS by industry subsector

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


fschreyer:
Hi Michaja, 
the naming convention with the "+" variables is not correct. 

You only need to insert a "+" for the variation in the last dimension. Pluses before you can drop. For example:

Industry Energy|++|Chemicals = Industry Energy|Chemicals |+|Biomass + Industry Energy|Chemicals |+|Fossil + Industry Energy|Chemicals |+|Synfuel

At the end, you should check whether your aggregation actually adds up by running the function [checkIntegrity](https://github.com/pik-piam/remind2/blob/75450c8b2ab38dd5e819002ebded67c2afb5bed2/tests/testthat/test-convGDX2mif.R#L29) on the output of ``convGDX2MIF``. You should check a policy run here to have carbon capture etc. in there. You can also check whether the right variables are aggregated by running mip::extractVariableGroups() on the names of the output of ``convGDX2MIF`` to see which variable groups are summed to which aggregate variable. 

I can add this documentation somewhere. Not sure where people usually look that up for an R-library.

amerfort (commented):
So following Felix' comment, the super-variables
Carbon Management|Carbon Capture|Industry Energy|Cement| ~|Steel| and ~|Chemicals| need to be added first, right? And I was told that the |++| denote an alternative way of summing up to the corresponding super-variable, so we wouldn't need it in this case here. Other than that I suggest to also include the new Industry variables on the Carbon Management|**Carbon Storage** side of the variable tree as well. If you drop me or @katarkow a line, we can include this, when we clean up the variable tree soon(ish).

fschreyer (approved):

fschreyer:
Ah, what about "other industry"? That does not capture carbon, right? In that case you can also give the sectors a "+": ``Carbon Management|Carbon Capture|Industry Energy|++|Cement (Mt CO2/yr)`` etc. as they should sum up to the total. The "++" you need because there is already an aggregation to ``Industry Energy`` with one plus (+|Biomass, +|Fossil).

PR #257, opened 2022-05-10
reporting subsidy or tax rate necessary to reach implicit set energy bounds

Renato-Rodrigues:


robertpietzcker (commented):
You can merge without my review, right?

my R capabilities are not sufficient to look at the code and see directly if there might be a problem/mistake hidden in it :-)

PR #258, opened 2022-05-13
make test-convGDX2mif.R more verbose, listing all elements of calculations that failed the integrity check

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Example output:
```
Failure (test-convGDX2mif.R:73:5): Test if REMIND reporting is produced as it should and check data integrity
Check on data integrity failed for Emi|GHG|Gross|Energy (Mt CO2eq/yr)
Emi|GHG|Gross|Energy|+|Demand (Mt CO2eq/yr)
Emi|GHG|Gross|Energy|+|Supply (Mt CO2eq/yr)
```

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '35763750'
+ValidationKey: '35782875'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


Loisel (approved):
Thanks for providing a more readable test output in the fail-case.

PR #266, opened 2022-05-19
Add variable 'Cap (GWel)|Hydrogen|Electricity' to capacity reporting

fbenke-pik:


fschreyer:
Hey Falk! Thanks! Could you also add the same for the ``New Cap|...`` variable [here](https://github.com/fbenke-pik/remind2/blob/e2cee0e6782dee85cffaf5c062829ab44d85d120/R/reportCapacity.R#L171) (so, the newly built GW(el)/yr in a time step)) This will also add it automatically for the ``Cumulative Cap|...`` variable). We should report all three in that unit if that is the new Ariadne definition.

Issue #254, opened 2022-05-10
No Documentation on |+| and |++| Mechanism


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Which makes it neigh impossible to work on the package without breaking something.

fschreyer:
Done via #271.

PR #268, opened 2022-05-20
Add documentation of development and variable naming guidelines to readme file

fschreyer:
addresses #254


giannou:
Cool, thanks for this. I was under the impression that the README file is generated automatically and therefore people where adding package documentation to a vignette rather than to the README. No clue how you managed to overcome this issue but if everything works fine then no problem from my side!

fschreyer:
Hi Anastasis, 

in fact I didn't build the library. You are right. Now, that I built it, I see that the ``readme.md`` is automatically overwritten. That's a pity. So, where would you suggest I put my documentation?

giannou:
I'd say let's write a vignette and add some in-code documentation also

PR #271, opened 2022-05-30
add documentation to remind2 package via vignette

fschreyer:
Adds documentation and development guidelines to the remind2 package. Addresses #254 . 

I simply added a vignette file to the library. However, I am not sure whether that is sufficient to make it show up as a vignette. Please let me know. Also, please correct, comment or extend (ideally make changes directly on github) where you find it necessary.


vignettes/remind2_reporting.Rmd
```diff
@@ -0,0 +1,92 @@
...
+You can check out the different summation groups of variables with "+" by running the function ``extractVariableGroups()`` from the [mip-package](https://github.com/pik-piam/mip#readme). The argument to the function is a vector of variable names. For example:
+
+```{r}
+mip::extractVariableGroups(getNames(testReport, dim=3))
+```
+
+
+The REMIND reporting will provide you with two different mif-files. The standard mif-file will come with "+" notation, while a second mif-file "without_plus" is generated where pluses were deleted to ease further processing. 
+
+Note that the defintions of REMIND variables as generated by the ``remind2`` library may not correspond to the definition of equally or similarly named variables in standardized model intercomparison projects. For mappings of REMIND variables to variable templates from model intercomparison projects, please see the pik-piam library [project_interfaces](https://github.com/pik-piam/project_interfaces).
```
orichters, line 91: definitions instead of defintions

vignettes/remind2_reporting.Rmd
```diff
@@ -0,0 +1,92 @@
...
+---
+
+```{r, echo = FALSE}
+knitr::opts_chunk$set(collapse = TRUE, comment = "#>")
+```
+
+
+## Function
+
+The ``remind2`` library contains the routines to generate the reporting of the output of the REMIND model and is the successor of the deprecated ``remind`` package since REMIND version 3.0. It's main function is to create the output of REMIND as a table in IAM-comparable format, the so-called ``mif-file``, from the ``fulldata.gdx`` of a run. The resulting ``mif-file`` is typically called ``REMIND_generic_NameOfYourRun.mif`` and can be found in the output folder of a run.
```
orichters, line 19: I think it must be "Its main function", not "It'**s** main function".

vignettes/remind2_reporting.Rmd
```diff
@@ -0,0 +1,92 @@
...
+testFE <- reportFE(gdx)
+```
+
+
+
+## Development
+
+
+
+When developing the remind2 library, please test your changes on a recent ``fulldata.gdx`` file from REMIND by running the ``convGDX2MIF()`` function. Make sure the ``fulldata.gdx`` file is suitable for testing the change you did. For example, check that the relevant GAMS variables and parameters are available and non-zero in the GDX you are testing. Different REMIND runs come with different module realization and switches where certain sets, parameters or variables are not existing such that your change may not be tested by an unsuitable GDX. Moreover, when adapating or adding variables with "+" notation, please run a summation check on a suitable GDX by running the ``checkIntegrity()`` found [here](https://github.com/pik-piam/remind2/blob/1a80cc607db663a211f98cabb3c7408cd08bf713/tests/testthat/test-convGDX2mif.R#L31) on your reporting output:
```
orichters, line 44: I took this as an incentive to write a PR that allows you to test other gdx files more easily: https://github.com/pik-piam/remind2/pull/272

vignettes/remind2_reporting.Rmd
```diff
@@ -0,0 +1,92 @@
...
+
+
+```{r}
+checkIntegrity(testReport)
+```
+
+
+If your change involves reading in a parameter, variable or set from the GDX-file that is not yet merged to the [main REMIND repository](https://github.com/remindmodel/remind) or that is not declared in the default runs of REMIND, please use conditional statements that check whether these parameters exist. Otherwise, the reporting will break for other users with other GDX files.
```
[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Not sure what you mean.  Anything read in that is not in [rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_fulldata.gdx](https://github.com/pik-piam/remind2/blob/1d3c4f98cb7e55d4769723b13da0a62db33b0894/tests/testthat/test-convGDX2mif.R#L58-L59) should throw an error and not even make it to the RSE package repository.

giannou (approved):

vignettes/remind2_reporting.Rmd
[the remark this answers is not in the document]
fschreyer: If the ``fulldata.gdx`` downloaded is very old, then it might not trigger the if-clauses of your run that is some weeks old. But, in general Michaja is right. 

The trade-off of a lot of if-clauses which never get cleaned-up is that the code becomes harder to read.

fschreyer:
@giannou: What about the vignette of remind2? How do I make it visible?

If I type ``?remind2``, nothing shows up which is different than in other packages. I guess, I'd still need to add something.

giannou:
The package man page and the vignettes should be visible now

orichters (approved):
All I could contribute was corrected, thanks, @fschreyer, for this effort! We should link this somewhere in https://github.com/remindmodel/development_issues/issues/101 so new users find it.

PR #272, opened 2022-05-30
add AMT tests

orichters:
This draft
- adds code such that a bunch of gdx files from AMT modeltests, the calibration gdx files and all in `/p/projects/remind/debugging/gdx-examples` can be added very quickly to the model tests to make testing various gdx files more easy.
- prints the filename with the groups that fail to facilitate debugging

Issue #66, opened 2021-07-07
Create log file to see how variable aggregations checks went


fschreyer:
It would be good to have a log file after running the aggregation test in ``test-convGDX2mif.R`` to see which aggregations have failed and which sum up as they should. This would

a) make sure the aggregation test checks all variables we want to check
b) make it easier to debug as currently only the parent variable affected is shown but not which aggregation (the one with one "+", two "+" etc.)

orichters:
done in https://github.com/pik-piam/remind2/pull/444

PR #274, opened 2022-05-31
adds variable summation check to convGDX2MIF

giannou:
separates summation check functions from package tests


giannou:
Now summation checks can be performed by simply running `convGDX2MIF`

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
Obvious point: the documentation is incomplete and mismatched.

On the usability: the idea is to factor out these functions from the tests, export them, and have users/scripts use them for testing and debugging.  Now, how do I generate the required data table?
I know Alois went with data tables because that is what he uses, which is fine for the internal tests.  But if REMIND people are expected to use this, the functions should accept either `magclass` objects (used throughout the package) and/or data frames (fine by me `;)`).

giannou:
> Obvious point: the documentation is incomplete and mismatched

Easy tiger. It's just a draft, I will not document functions that will be rejected.

orichters:
You may want to have a look at those changes I added in https://github.com/pik-piam/remind2/pull/272/ that are related to `gdxPath`. If you can apply the checks to a file, I would appreciate if there was at least the possibility to pass the path as a string such that the error message was more readable, if one checks more than one file.

giannou:
> Then why is `checkEqs()` exported?

I exported it while testing stuff during implementation. Can be undone easily.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I exported it while testing stuff during implementation. Can be undone easily. 

Bit early for review requests, maybe.

> Also, I'm not sure if the `testConvGDX2MIF.R` will see it if not exported.

You can always call via `:::`.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
As per #66, `checkIntegrity()` could return the data for which summation fails, for debugging.

giannou:
> Bit early for review requests, maybe.

It's a collaborative effort and the easiest way to do this is to discuss it here.

fschreyer:
That's a good step! Thanks. 

> Anybody wants to touch Alois' data table code, or will we just rewrite it?

I think converting magclass input to data.table format in the beginning might also be sufficient.

giannou:
> converting magclass input to data.table format in the beginning

could you elaborate a bit on that?

Loisel:
I have no strong opinion. I can of course modify the dt code to your liking, but I agree that code with less of a hurdle for the majority is preferable. I also think having such a function exported and interfacing `quitte` or `magclass` is a good idea (this, of course, could also be done with the existing code and some wrapper).

giannou:
I added a quitte-based summation check function `checkSummations`. Example output:
```
> checkSummations(x)
[[1]]
[1] "failing and/or missing variable groups"

[[2]]
 [1] "FE|Solids|+|Biomass"                 "FE|Solids|+|Fossil"                  "FE|Industry|Chemicals|+|Electricity" "Emi|CO2|+|Energy"                   
 [5] "Emi|CO2|Gross|Energy"                "Emi|GHG|+++|Energy"                  "Emi|GHG|Gross|Energy"                "Emi|GHG|++|ETS"                     
 [9] "Revenue|Government|Tax|Carbon"       "FE|Industry|+++|Steel"  
```

Loisel:
I vaguely remember that for the old version one requirement was that one can add non-`+`-notation based summation checks, I think @fschreyer was also part of the discussion back then.

giannou:
Summary:

- the function checkSummations(data) runs summations checks based on the `|+|` notation and returns a quitte with the failed data (an additional column `parent` exists to help debugging)
- you can perform checks on additional variables (without the `|+|` notation) by adding them in the file `summationGroups.csv` in `inst/extdata`
- the summation checks are included in `convGDX2MIF`, which means they are performed at the end of every REMIND run that has `reporting` as output

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> - you can perform checks on additional variables (without the `|+|` notation) by adding them in the file `summationGroups.csv` in `inst/extdata`

So we've come full circle from deciding to integrate this summation information in the data itself, instead of having a dedicated file, to then adding code everywhere to get rid of this extra information, to now having both -- funky plus signs in the data and a dedicated file to list out summations.

giannou:
@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q feel free to suggest something better. Removing the `|+|` notation is unrealistic, so make sure to keep that in mind in your suggestion,

orichters:
> @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q feel free to suggest something better. Removing the `|+|` notation is unrealistic, so make sure to keep that in mind in your suggestion,

Do we need the additional possibility using `inst/extdata/summationGroups.csv`? Why is the `|+|` notation not sufficient?

giannou:
> Do we need the additional possibility using `inst/extdata/summationGroups.csv`? Why is the `|+|` notation not sufficient?

Doesn't hurt to have it if someone needs it, does it? For example, someone might want to quickly check something in their data. @Loisel and @fschreyer can say more on that.

fschreyer:
> > @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q feel free to suggest something better. Removing the `|+|` notation is unrealistic, so make sure to keep that in mind in your suggestion,
> 
> Do we need the additional possibility using `inst/extdata/summationGroups.csv`? Why is the `|+|` notation not sufficient?

I think, it is quite useful to have the summation group file in addition because sometimes you wanna report/check summations which do not fit into this "+" naming scheme. For example, you might wanna check the summation of a group of emissions variables with bunkers, e.g.  ``Emi|CO2|w/ Bunkers`` = ``Emi|CO2`` + ``Emi|CO2|Energy|Demand|Transport|International Bunkers``. So I like this option.

Renato-Rodrigues:
I second the necessity for something more than the plus only check for examples like Felix mentioned.

I would prefer although to defined those in a single place only.
 
As for more complex scenarios, ex: using the same variable for multiple summation checks, is not possible using the sum signs anyway, I would vote for it to be deprecated some time in the future in favor of using a single solution using am external file to define all summation checks.
iiasa does/has plans to do that way in their variables template definitions, and this would also avoid the duplicated mif file output that we currently create.

orichters:
What is the state of this project? I still think it would be very useful to have it…

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Pretty dead I guess, with Anastasis and Alois gone.

fbenke-pik:
Who is still waiting for this to be merged? I can look into it, but would need someone to bring me up to speed. From the discussions above, I don't feel like I have a clear understanding what should be achieved.

From RSE side, I don't think we have anything related to this on our immediate TODO list, so reach out to me if you'd like to put this back on our agenda.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> From RSE side, I don't think we have anything related to this on our immediate TODO list, so reach out to me if you'd like to put this back on our agenda.

Strange.  I could swear I wrote a note tagging you and asking you to bring this up with RSE to decide whether we still want this or not.

Stupid internet.

> I can look into it, but would need someone to bring me up to speed. From the discussions above, I don't feel like I have a clear understanding what should be achieved.

The general idea is to 
1. have the summation checks available outside of the unit tests, to test specific data
2. extend the summation checks from the `|+|` syntax to include specific mappings (say for analysing a bunch of scenarios submitted by project partners)

orichters:
I also would appreciate if something like the [checkIntegrity](https://github.com/pik-piam/remind2/blob/master/tests/testthat/test-convGDX2mif.R#L37) function would be made available outside of the tests, improved and expanded.

Concerning "extend the summation checks from the |+| syntax to include specific mappings (say for analyzing a bunch of scenarios submitted by project partners)" as proposed by @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, I think that should not be part of `remind2` package, as they do not tend to submit in `remind2` output format, but rather using the `AR6` or `NAVIGATE` template. And for this use case, we have [piamInterfaces::checkSummations()](https://github.com/pik-piam/piamInterfaces/blob/master/R/checkSummations.R) and [checkSummationsRegional()](https://github.com/pik-piam/piamInterfaces/blob/master/R/checkSummationsRegional.R) already and I think that is the place where improvements should be implemented, if needed.

fbenke-pik:
I will look into consolidation at some point in July. Probably will open a new PR/issue, as this was done before we had mip. Will keep you posted.

fbenke-pik:
I looked into how to consolidate Anastasis' work with piamInterfaces::checkSummations as well as the quitte option. 

Option 1)
In order to reuse piamInterfaces, we just need to extend checkSummations to be able to extract the summations group from variable names. See this [draft PR](https://github.com/pik-piam/piamInterfaces/pull/146/files).

Option 2)
In order to use the quitte function, we would need to extend it as well to support variable extraction (currently it expects a text file with summations or a list of summations). It also needs support of two or more alternative summations for the same variable. 

If you want to go with Option 1, i can implement this. For Option 2, Michaja should probably do it.

In any case, I don't think it is worth our time to further consolidate `piamInterfaces::checkSummations()` and `quitte::check_quitte()`. I think both have different use cases and therefore should exist alongside. 

Let me know how I should proceed. If there are no objections, I will go ahead with Option 1 and write a separate test case in remind2 calling piamInterfaces::checkSumations.

orichters:
If the draft PR for `piamInterfaces` works, I have no objection to using it. Option 1B does not seem more convincing / easy to me.

fbenke-pik:
Closing in favour of https://github.com/pik-piam/remind2/pull/444
Please reopen or create a new issue, if you feel like this PR contains features that have not been addressed by this PR.

PR #275, opened 2022-05-31
use Emi|Kyoto Gases for Revenue reporting

orichters:
as discussed within NGFS


christophbertram (approved):
This looks good, thanks

PR #276, opened 2022-06-02
add optional parameter modelHistExclude to compareScenarios2.R

fbenke-pik:
The optional parameter modelHistExclude allows removal of models from historical data. In some contexts, some historical data is not useful, for example does is not make sense to plot IEA ETP in plots with focus on Germany because it is coarsely downscaled from EU data and therefore not meaningful.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '36373600'
+ValidationKey: '36394645'
 AcceptedWarnings:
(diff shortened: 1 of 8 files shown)


chroetz (approved):
works without problems.

Note: If a historical model is excluded, it will still show up as a level of `data$model` even so there is no row with that model anymore. One could add `data <- droplevels(data)` to clean this up. But it is not strictly necessary as the plot functions will call `droplevels()` anyway.

PR #277, opened 2022-06-03
adds comparison of steel accumulation and gdp percapita

mellamoSimon:
hi there!
I added some plots to compareScenario2 to help in the analysis of the relation between GDP and the accumulation of steel in different regions.

### DIFF
--- a/inst/markdown/compareScenarios2/cs2_06_energy_services.Rmd
+++ b/inst/markdown/compareScenarios2/cs2_06_energy_services.Rmd
@@ -206,10 +206,36 @@ stock <- data %>%
 showLinePlots(stock, "Production|Industry|Steel|Accumulated Production", mainReg = "World")
 
 stock_pc <- left_join(stock, population, by = c("period", "region", "scenario")) %>%
-  mutate(value = value.x / value.y, unit = "t per cap")
+  mutate(value = value.x / value.y, unit = "t per cap") %>% 
+  select(-value.x,-value.y)
 
 showLinePlots(stock_pc, "Production|Industry|Steel|Accumulated Production", scales = "fixed", mainReg = "World")
 ```
+
+###Steel accumulation versus GDP evolution and rate 
+```{r}
+steelGDP <- data %>%
+  filter(variable == "GDP|PPP pCap") %>%
+  full_join(stock_pc, by = c("region","scenario","model","period")) %>%
+  mutate(value = value.y/value.x) %>%
+  mutate(variable = "Production|Industry|Steel|Accumulated Production per GDP PPP", unit = "t_pc/ GDPpc") %>%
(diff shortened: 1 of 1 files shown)


inst/markdown/compareScenarios2/cs2_06_energy_services.Rmd
```diff
@@ -206,11 +206,37 @@ stock <- data %>%
 showLinePlots(stock, "Production|Industry|Steel|Accumulated Production", mainReg = "World")
```
chroetz, line 206: Sorry, I missed that in your last PR, but its better to not explicitly set `mainReg = "World"`. If you just remove this argument (here and in your calls to `showLinePlots()`) below, the function will use the default value for the main region. The default is `getOption("mip.mainReg")` which is globally set to the correct main region. This could be "World" but also e.g. "EUR".

chroetz (changes requested):

inst/markdown/compareScenarios2/cs2_06_energy_services.Rmd
```diff
@@ -206,11 +206,37 @@ stock <- data %>%
...
 
 stock_pc <- left_join(stock, population, by = c("period", "region", "scenario")) %>%
-  mutate(value = value.x / value.y, unit = "t per cap")
+  mutate(value = value.x / value.y, unit = "t per cap") %>% 
+  select(-value.x,-value.y)
 
 showLinePlots(stock_pc, "Production|Industry|Steel|Accumulated Production", scales = "fixed", mainReg = "World")
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 212: I would go with "cumulated production".  "Accumulated" is too suggesting too strongly that this is the accumulation of steel in the economy, for which the depreciation is missing.

inst/markdown/compareScenarios2/cs2_06_energy_services.Rmd
```diff
@@ -206,11 +206,37 @@ stock <- data %>%
...
 ```
 
+###Steel accumulation versus GDP evolution and rate 
+```{r}
+steelGDP <- data %>%
+  filter(variable == "GDP|PPP pCap") %>%
+  full_join(stock_pc, by = c("region","scenario","model","period")) %>%
+  mutate(value = value.y/value.x) %>%
+  mutate(variable = "Production|Industry|Steel|Accumulated Production per GDP PPP", unit = "t_pc/ GDPpc") %>%
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 221: See above.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

PR #281, opened 2022-06-10
improve compareScenConf description

orichters:
to make it more evident which the old and which the new file is

Also, don't print `(default: )` for the `description`


mikapfl (approved):
Is clearer, thanks!

orichters:
Sorry, I did another small tweak: I changed the part behind again, so that the ` (default: )` will not be written if there is no value in `cfg$gms`.

mikapfl (approved):

PR #285, opened 2022-06-22
update workflows

orichters:
It is the result of running `lucode2::buildLibrary()`, but I first couldn't submit it because of issues @mikapfl could fix. I'm not sure whether I have the right to do that (edit: seems to work)


Renato-Rodrigues:
You should probable ask somebody from RSE to revise this PR. Currently, my knowledge on the lucode2 backbone is very limited.

pfuehrlich-pik (approved):

PR #286, opened 2022-06-23
* reporting FE marginal prices directly from the model and fixing transport complex report

Renato-Rodrigues:
* adding FE marginal prices reporting directly from the model results
* fixing compatibility issues with transport complex

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '364421700'
+ValidationKey: '364440870'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.90.10",
+  "version": "1.90.11",
   "description": "<p>Contains the REMIND-specific routines for data and model output manipulation.<\/p>",
   "creators": [
     {
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: remind2
 Type: Package
 Title: The REMIND R package (2nd generation)
-Version: 1.90.10
+Version: 1.90.11
 Date: 2022-06-27
 Authors@R: as.person(c(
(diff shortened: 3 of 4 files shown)


fschreyer (changes requested):
Cool! You managed to extract FE prices by SE origin. Curious to see how you did that and what the results look like in the end. 

Why do you want to call them "marginal" prices and add them in addition to the standard prices? The other prices are also based on marginal values of the model. Could we maybe replace the prices we have with them?

It would be good to have an explanation (either here or in the GAMS code) in what way they differ from the standard prices.

Renato-Rodrigues:
For knowing how I implemented it, you can see the code in this PR (unfortunately it is full with other interlinked changes):
https://github.com/remindmodel/remind/pull/848
Just check the end of the core/postsolve file and the equation qm_balFeAfterTax to see how do I get those values.

I called them "PRICE|MARGINAL|" because so far there was a lot of misunderstanding of people not getting that marginals can be zero if an equation is not active, so I wanted to be more explicit about that.

Once the REMIND PR is merged, some work could be done to use these marginal prices in the reporting instead of the ad-hoc disaggregations and weights used right now that are much more susceptible to bugs and mistakes. This should simplify substantially our prices report.

I would also advocate in favor of cleaning up our "PRICE|FE" variables, treating them so we can create project reportable prices that avoid the usual issues, ex: the "this price is zero" complain that comes again and again from other people that do not understand marginal prices; complains about fluctuations between time steps;...

I am not sure right now if it is necessary to change the "MARGINAL" string with something like "GAMS-Marginals". I could do it in the future if necessary, but I tend to like smaller names better. The reporting variables can have quite long names already, and I would hold on any renaming for now before we have a final overview of all possible price variable variations that we would want to add to the report.
For example, nothing avoids us to report prices before CO2 and other taxes are applied. I plan to do that someday, but the structure is in place to do it already.
We shoudl stop for a few minutes in the future to make a list of the desirable disaggregations that we could potentially want before doing any definitive renaming in my opnion.

fschreyer (approved):

fschreyer:
Ok, I don't fully understand but I see that some prices which were aggregated before are now directly taken from GAMS marginals.

Well, let's introduce these new prices with "|Marginal|", look at results and maybe replace out our old price variables later on.

Issue #287, opened 2022-06-23
Capital Costs incorrectly aggregated


orichters:
There is a strange bug in the aggregation of Capital Costs (a thing that I reintroduced yesterday in #284), see the mif file
```
REMIND;C_h_ndc_bIT-rem-5;USA;Tech|Electricity|Biomass|Gasification Combined Cycle w/ CC|Capital Costs;US$2005/kW;7007.7386498;7007.7386498;6703.0543606;5484.3172042;4570.2643368;3656.2114694;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;3046.8428912;
REMIND;C_h_ndc_bIT-rem-5;World;Tech|Electricity|Biomass|Gasification Combined Cycle w/ CC|Capital Costs;US$2005/kW;89370.468733;89370.468733;85484.7961793;5485.0721023;4571.9359698;3657.6323218;3048.3504831;3048.7390548;3048.7546132;3048.783714;3048.8354978;3048.9650924;3050.6720627;3238.2045504;3238.0604613;3240.6866848;3250.353158;38856.7255361;38856.7255361;
```
You see that in the first three and the last two periods of `World`, the sum is calculated, while in the other time steps, the average is calculated as it should be. I have no clue why.

Renato-Rodrigues:
Variables that could have weights equal to zero should not be used for price/cost aggregations. Is the same issue that we had with CO2 prices aggregations. You should use instead some weight variable that does not include zero values to avoid this artifact, or simply add a filter before to the weighted aggregations (before here: https://github.com/pik-piam/remind2/blob/6813192b8881feb885b55b983ca038ef777c3a18/R/reportTechnology.R#L361), that remove the variables that have zero weights from the regional calculated aggregations.

Renato-Rodrigues:
Could you share the values for the variable, and the weights used on this specific case so we could see what is the actual behavior.

orichters:
Here is the test: The problem are years in which no single region has above zero values as a weight. Then [speed_aggregate](https://github.com/pik-piam/luscale/blob/master/R/speed_aggregate.R#L64) returns the sum.
```
library(gdx)
library(magclass)
library(luscale)
library(remind2)
[...]

PR #289, opened 2022-06-24
fix the aggregation issue for capital costs

orichters:
This is a maybe not entirely beautiful fix to https://github.com/pik-piam/remind2/issues/287.

If a certain technology is not used in any region, `speed_aggregate` returned the sum of the prices in [this line of reportTechnology.R](https://github.com/orichters/remind2/blob/a33b35a912a21d211bedbfe0c6692aa0d80d9362/R/reportTechnology.R#L363) instead of the weighted average value.

Setting each value to a minimum of `1E-30` fixes this problem. The values in `output` for all these variables are `≥ 0`, so we don't cut negative values here, and the smallest non-zero value derived from my GDX is of the order of `E-12`, so setting everything on a minimum of `1E-30` only in this aggregation process does not affect actual values.

I checked that it only has impact on the value of the variables that were wrong, and mostly affect 2005, 2010, 2015, 2110, 2130 and 2150 values, except for Hydrogen were also 2020 and 2025 values were affected.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '36570636'
+ValidationKey: '36589803'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.90.8",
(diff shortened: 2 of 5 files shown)


Renato-Rodrigues (approved):

orichters:
Renato, I checked [Jans suggestion](https://github.com/pik-piam/luscale/issues/1) to use `toolAggregate`, but it doesn't solve the problem, returning not the sum, but just `0` in case of all `0` weights, which isn't what we want. So I merge, we can improve it later.

PR #299, opened 2022-07-07
Add sub-categories of landuse emissions to REMIND report

dklein-pik:
They seem to have vanished when the emisssions reporting was refactored.


merfort (approved):
Thanks! Looks good to me

PR #302, opened 2022-07-19
fix linter issues for cs2

chroetz:


inst/markdown/compareScenarios2/cs2_99_further_info.Rmd
```diff
@@ -24,29 +24,30 @@ showInTables <- function(
...
           latex_options = c("striped", "hold_position")
-        ) ->
-        out
+        )
     } else {
-      out %>%
+      out <-
+        out %>%
         column_spec(
           1:(nVarCols + 1),
```
pfuehrlich-pik, line 41:
```suggestion
          seq_len(nVarCols + 1),
```

inst/markdown/compareScenarios2/cs2_01_summary.Rmd
```diff
@@ -4,7 +4,7 @@
 
 ## GHG Emissions
 
-```{r Emi-GHG}
```
pfuehrlich-pik, line 7: Did the linter complain about this? Or did you just remove because it is kinda redundant?

pfuehrlich-pik (approved):
Thanks for this! Looks like a lot of not-so-interesting work. I hope the auto-formatter did most of it :sweat_smile:

inst/markdown/compareScenarios2/cs2_01_summary.Rmd
chroetz: No, but they are redundant and if you happen to label two sections with the same name, everything crashes. This can easily happen if you split a chunk via shortcut; then both resulting chunks have the same name. So better to not use labels for chunks unless they are actually useful.

chroetz:
It was pure fun 🙃

PR #308, opened 2022-07-26
prepare reportSE for all_enty split

chroetz:
`emi` will be split form `all_enty` in REMIND. Then `pm_prodCouple` will not contain `ccsinje` anymore. Instead this will be in `pm_prodCoupleEmi`. 

This change allows both situations (old and future) to be handled.


robertpietzcker:
> Then `pm_prodCoupleEmi` will not contain `ccsinje` anymore. Instead this will be in `pm_prodCoupleEmi`.

stupid question: is this a typo? to me both "pm_xxx" look the same?

R/reportSE.R
```diff
@@ -61,13 +61,19 @@ reportSE <- function(gdx, regionSubsetList = NULL, t = c(seq(2005, 2060, 5), seq
 
   ## parameter
   pm_prodCouple    <- readGDX(gdx, c("pm_prodCouple", "p_prodCouple", "p_dataoc"), restore_zeros = FALSE, format = "first_found")
+
   pm_prodCouple[is.na(pm_prodCouple)] <- 0
+  teprodCoupleSeel <- getNames(mselect(pm_prodCouple, all_enty2 = "seel"), dim = 3)
+  if (!"ccsinje" %in% teprodCoupleSeel) {
+    pm_prodCoupleEmi <- readGDX(gdx, c("pm_prodCoupleEmi"), restore_zeros = FALSE, format = "first_found")
+    pm_prodCoupleEmi[is.na(pm_prodCoupleEmi)] <- 0
```
LaviniaBaumstark, line 69: looks a bit strange that adding this set with value =0 is equal to the old version where "ccsinje" was included... Was ist always 0? Or should another parameter be used now?

Renato-Rodrigues:
I think it should read:

Then `pm_prodCouple` will not contain `ccsinje` anymore. Instead this will be in `pm_prodCoupleEmi`.

R/reportSE.R
Renato-Rodrigues: I also don't think this will work.

As far as I understand, you just need to test if you have pm_prodCoupleEmi in the model, in positive case (new formulation) you just load its values from the gdx, in negative case (old formulation) you simply assume its values equal to pm_prodCouple[,,ccinje], no?

```
pm_prodCoupleEmi <- readGDX(gdx, c("pm_prodCoupleEmi"), restore_zeros = FALSE, format = "first_found")
if (!(is.null(pm_prodCoupleEmi))){
pm_prodCouple[,ccinje] <- pm_prodCoupleEmi 
}

This way you wouldn't need to change anything else in the code, no?

**Review by Renato-Rodrigues — CHANGES_REQUESTED** [review]

**Comment by chroetz** [comment]
> > Then `pm_prodCoupleEmi` will not contain `ccsinje` anymore. Instead this will be in `pm_prodCoupleEmi`.
> 
> stupid question: is this a typo? to me both "pm_xxx" look the same?

Yes, it's a typo. I corrected it

File: R/reportSE.R (hunk 1 above)
Comment (chroetz) [reply R/reportSE.R:69]: @LaviniaBaumstark sorry, I don't really understand your comment. What do you mean by "adding this set with value =0"?

Comment (chroetz) [reply R/reportSE.R]: @Renato-Rodrigues You are right! Less change in the code is required if `pm_prodCoupleEmi` is just added back to `pm_prodCoupleEmi` in case it is separated. I commited a version of your code draft which uses `mbind` to achieve that. By that I don't even have to check with `is.null()` as `mbind(x, NULL)` is the same as `x`.

File: R/reportSE.R [hunk 2]
```diff
@@ -60,14 +60,17 @@ reportSE <- function(gdx, regionSubsetList = NULL, t = c(seq(2005, 2060, 5), seq
   input_liquids <- c(entyPe, "seh2")
 
   ## parameter
-  pm_prodCouple    <- readGDX(gdx, c("pm_prodCouple", "p_prodCouple", "p_dataoc"), restore_zeros = FALSE, format = "first_found")
+  pm_prodCouple <- readGDX(gdx, c("pm_prodCouple", "p_prodCouple", "p_dataoc"), restore_zeros = FALSE, format = "first_found")
+  pm_prodCoupleEmi <- readGDX(gdx, "pm_prodCoupleEmi", restore_zeros = FALSE, react = "silent")
```
LaviniaBaumstark, line 64: does pm_prodCoupleEmi also exist in the old version?

chroetz: no. `readGDX` will return `NULL` (without warning because of `react = "silent"`)

Renato-Rodrigues (approved):
As long as the code is tested with an old and new gdx I don't have any additional comments.

By the way, thk you for sharing the trick to avoid the need for the conditional. I knew in the back of my mind that NULLs are disregarded but I always used conditional by some reason. You did a much more elegant solution. I will definitely use that in the future. :)

LaviniaBaumstark (approved):

PR #313, opened 2022-08-02
Update cs2

robertpietzcker:
cs2: corrected Investments Electricity area plot
cs2: added SE line plots


chroetz (approved):
Tested it and works without problem

PR #316, opened 2022-08-02
adds new function `getCs2Profiles()`

chroetz:
Adds a function to load the cs2-profiles from the REMIND folder. Will  be used in https://github.com/remindmodel/remind/pull/926


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

R/getCs2Profiles.R
```diff
@@ -0,0 +1,27 @@
...
+#' profiles <- getCs2Profiles() # load from default location
+#' profiles <- getCs2Profiles(remindDir = "path/to/remind")
+#' profiles <- getCs2Profiles(path = "path/to/profiles.json")
+#' profiles <- getCs2Profiles(relativePath = "path/relative/to/remindDir/profiles.json")
+#' }
+#' @export
+getCs2Profiles <- function(
+    path = file.path(remindDir, relativePath),
+    relativePath = "scripts/cs2/profiles.json",
+    remindDir = ".") {
```
mikapfl, line 21: Kind of complicated with three arguments. If I specify for example `path` and `remindDir`, `remindDir` is simply ignored, right?
A simpler API would be to just have `path = "./scripts/cs2/profiles.json"` and let users of the function deal with defining the path themselves if they only know a `remindDir`.
Or, if only knowing the `remindDir` is common for users of the function, and we don't want to duplicate `scripts/cs2/profiles.json` in every user, maybe it would also be fine to have `path = "./scripts/cs2/profile.json"` by default and then allow the path to be a directory?
Something like
```
if (dir.exists(path)) {
    path = file.path(path, "scripts", "cs2", "profiles.json")
}
profiles <- jsonlite::read_json(path …)
```

mikapfl (approved):
Looks good to me, I'd maybe simplify the three function arguments to only one.

R/getCs2Profiles.R
chroetz: You are right. This is a bit overkill for a function that will mostly be called with its default argument... I will change it to your first suggestion

PR #318, opened 2022-08-05
reportCapacity: correct and extend heat capacity reporting

robertpietzcker:
+ added CHP capacities to heat capacities (previously only hp capacities were reported, now also chp)
+ geohe refers to Heat Pump, not Geothermal

### DIFF
--- a/.github/workflows/lucode2-check.yaml
+++ b/.github/workflows/lucode2-check.yaml
@@ -11,7 +11,7 @@ on:
 env:
   USE_BSPM: "true"
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  NO_BINARY_INSTALL_R_PACKAGES: 'c("madrat", "magclass", "citation", "gms", "goxygen")'
+  NO_BINARY_INSTALL_R_PACKAGES: 'c("madrat", "magclass", "citation", "gms", "goxygen", "GDPuc")'
 
 jobs:
   lucode2-check:
@@ -56,6 +56,11 @@ jobs:
 
       - uses: r-lib/actions/setup-pandoc@v1
 
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
       - name: Cache R libraries
         uses: pat-s/always-upload-cache@v3
         with:
@@ -73,6 +78,11 @@ jobs:
           ./run.sh install_aptget libhdf5-dev libharfbuzz-dev libfribidi-dev
           ./run.sh install_all
           ./run.sh install_r lucode2 covr rstudioapi
+
+      - name: Install python dependencies if applicable
+        run: |
+          [ -f requirements.txt ] && python -m pip install --upgrade pip wheel || true
+          [ -f requirements.txt ] && pip install -r requirements.txt || true
 
       - name: Remove bspm integration # to get rid of error when running install.packages
         run: |
--- a/.github/workflows/lucode2-check.yaml
+++ b/.github/workflows/lucode2-check.yaml
@@ -11,7 +11,7 @@ on:
 env:
   USE_BSPM: "true"
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  NO_BINARY_INSTALL_R_PACKAGES: 'c("madrat", "magclass", "citation", "gms", "goxygen", "GDPuc")'
+  NO_BINARY_INSTALL_R_PACKAGES: 'c("madrat", "magclass", "citation", "gms", "goxygen")'
 
 jobs:
   lucode2-check:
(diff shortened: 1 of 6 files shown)


.github/workflows/lucode2-check.yaml
```diff
@@ -11,7 +11,7 @@ on:
 env:
   USE_BSPM: "true"
   _R_CHECK_FORCE_SUGGESTS_: "false"
-  NO_BINARY_INSTALL_R_PACKAGES: 'c("madrat", "magclass", "citation", "gms", "goxygen")'
+  NO_BINARY_INSTALL_R_PACKAGES: 'c("madrat", "magclass", "citation", "gms", "goxygen", "GDPuc")'
```
Renato-Rodrigues, line 14: do you need the GDPuc reference here?

.github/workflows/lucode2-check.yaml
```diff
@@ -56,6 +56,11 @@ jobs:
 
       - uses: r-lib/actions/setup-pandoc@v1
 
+      - name: Set up Python 3.9
```
Renato-Rodrigues, line 59: I don't see any Python code added, why would you need to add a Python reference here? Was this added automatically by some RSE code? If yes, could you double check with them if this change is necessary.

.zenodo.json
```diff
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.99.0",
+  "version": "1.100.0",
```
Renato-Rodrigues, line 3: IT is not a big deal, but for changes like this we don't add to the version at this level, the recommended approach would be to use "1.99.1" instead.

Renato-Rodrigues (changes requested):

.github/workflows/lucode2-check.yaml
robertpietzcker: no idea how that was added?
It was no change that I did by hand, but rather the outcome of lucode2::buildLibrary() 

if you have an idea how I could remove it, I will be happy to do so.

.github/workflows/lucode2-check.yaml
robertpietzcker: same as above - no idea why or how that was added. 
I put a question in the RSE channel

.zenodo.json
robertpietzcker: ok, I interpreted the descriptions in the list that shows up when building a library such that this level should be increased when adding variables (like the CHP variables for electricity I added).

what would be example cases when you would increase this level?

.github/workflows/lucode2-check.yaml
fbenke-pik: I personally think you don't need to worry about changes like these, as long as you use the latest versions of libraries when running buildLibrary. They were introduced by running buildLibrary, so you can assume that RSE intended this.

.zenodo.json
fbenke-pik: I agree with Robert, he increased according to the description buildLibrary provides. 3rd digit increase is for bugfixes, 2nd digit increase is for new features. @Renato-Rodrigues if you feel like this is not the right way to do it, you should bring it up with RSE and ask for an adjustment of the recommendations the library gives you.

robertpietzcker: I mean "new features" is a very vague term - so if Renato says "simply adding extensions of existing variables "cap|elect|CHP" doesn't count as new feature - for a new feature I would have to add a new class of variables such as 'idle capacity' or so" I am totally fine with that and will do so in the future. I just wanted to understand in what case I should increase 2nd digit and in what case 3rd digit.

fbenke-pik: Don't get me wrong, i can see why this is a 3rd digit increase, not a 2nd digit increase and would myself also choose 3rd digit increase. But due to the vague description, people are left wondering and this will keep happening. So i can see why you came to that conclusion using the description.

.github/workflows/lucode2-check.yaml
Renato-Rodrigues: I would ask somebody from the RSE group about that (you can use the mattermost channel). 
Adding strict dependencies is not desirable in any library. If this is a new library requirement demanded by other code change, it should be added in a PR that explain why it is needed.
As your code does not require GDPuc in any way, I would simply remove this change from your PR, and commit only the removal without the need to update any version number.

.zenodo.json
Renato-Rodrigues: It was RSE that requested me to not use the subversion (2nd digit) for smaller code changes. So if you want to change the description there you should contact them directly.
They said to me that the second digit is reserved to more complex, long and/or breaking changes. Smaller changes such as small number of variables introduced or small bug fixes should only update the third digit instead.

.github/workflows/lucode2-check.yaml
Renato-Rodrigues: again, I would contact RSE to create a PR that includes only these changes and the justification why they were added.
IN your case I would simply edit those files by hand and remove the changes that are not related with your PR.

Renato-Rodrigues (approved):

PR #321, opened 2022-08-10
use gms::getLine()

orichters:
instead of defining it again in remind2. Part of https://github.com/remindmodel/remind/issues/929


mikapfl (approved):
Thanks!

PR #324, opened 2022-08-15
Adds function createVarListHtml()

chroetz:
* Makes `loadModeltest()` more robust.
* Adds new function `createVarListHtml()`. 
  * It writes an HTML document of a hierarchical list of variables extracted from mif-data. 
  * The HTML document is standalone. It does not use JavaScript - only HTML and CSS. The folding of list elements is made possible by the HTML5-tag `<details>`. This tag may not work with very old browsers like *Internet Explorer*, but *Firefox*, *Chrome*, and *Edge* are fine.
  * It is based on `variablesAsList()` and can display all the information that is in the result of a call to `variablesAsList()` with argument `entry = "INFO"`. Also its new `details` argument may be used, see below.
* `variablesAsList()` gains new argument `usePlus` to control whether variable names should be parsed with or without `|+|`.
* `variablesAsList()` gains new argument `details` to put further information into the INFO nodes of the hierarchical list it creates.

**Example 1**
```
library(remind2)
loadModeltest() # creates object data in global environment
createVarListHtml(data, "variables.html")
```
The result is `/p/tmp/cschoetz/variables.html`.

**Example 2**
```
library(remind2)
loadModeltest()
detailsAR6 <-
  readr::read_delim(
    "https://raw.githubusercontent.com/pik-piam/project_interfaces/master/ar6/mapping_template_AR6.csv",
    delim = ";",
    col_select = c(r21m42, Definition)
  ) %>%
  rename(name = r21m42)
createVarListHtml(
  data,
  "variablesWithDescription.html",
  title = "Reported REMIND Variables with AR6 Description",
  usePlus = TRUE,
  details = detailsAR6)
```
The result is `/p/tmp/cschoetz/variablesWithDescription.html`.


fschreyer:
Thanks for this! Could you maybe append an example html?

chroetz:
> Could you maybe append an example html?

Sure!  Here is an example: `/p/tmp/cschoetz/variables.html`
This is the result of running the code mentioned in the description.

fschreyer:
Looks goot to me! Thanks, very useful! Also maybe to throw out things at some point. There are a couple of things I have never seen. 

Do you plan to write this html to the remind2 repository? In my opinion, that would be the most visible spot. Or we ask RSE about that as it is about ``buildLibrary()``. 

Just food for thought if you have time to coordinate this: Dream would be ofc to append variable definitions in this html. A start might be to join the ``r21m42`` and ``Definition`` columns from the [AR6 mapping](https://github.com/pik-piam/project_interfaces/blob/master/ar6/mapping_template_AR6.csv) for REMIND and add this as a definition file in remind2, for example. Still, details would need to be figured out and aligned afterwards in quite a team effort which should probably be coordinated with RSE.

chroetz:
> Do you plan to write this html to the remind2 repository? In my opinion, that would be the most visible spot. Or we ask RSE about that as it would need to be integrated in buildLibrary().

I'm not sure, whether we need to create the HTML every time `buildLibrary()` is called. Maybe we could also just run it every time the automated model test have completed? But I don't have a strong opinion on that.

> Just food for thought if you have time to coordinate this: Dream would be ofc to append variable definitions in this html. A start might be to join the r21m42 and Definition columns from the [AR6 mapping](https://github.com/pik-piam/project_interfaces/blob/master/ar6/mapping_template_AR6.csv) for REMIND and add this as a definition file in remind2, for example. Still, details would need to be figured out and aligned afterwards in quite a team effort which should probably be coordinated with RSE.

I added an argument `details`. It takes a data frame with a column `name` and further columns. The information provided in these columns is appended to the INFO nodes of variables with the name `name`. See the new second example in the description above.

fschreyer (approved):

PR #325, opened 2022-08-16
add various variables related to CES function in reportMacro for diagnostics and add corresponding plots to cs2

fschreyer:
This adds the following sets of variables related to the CES function:

* ``Internal|CES Function|CES Price|input`` -> gives CES price (CES derivative) of input
* ``Internal|CES Function|MRS|input1|input2`` -> gives marginal rate of subsitution of input1 w.r.t to input2
* ``Internal|CES Function|Value Added|Input`` -> gives value added (price * quantity) of input

All variables start with ``internal`` as they should usually not be reported to external projects but rather be used for internal diagnostics to better understand model results. For external users, make sure they understand the meaning of the variables. 

Furthermore, a corresponding section of CES function reporting is added in compare scenarios 2. It is on by default. However, as its output is quite specific one could think about making this section optional for users. 

Example cs2 plot are here:

``/p/tmp/schreyer/Modeling/remind/ElecH2_Paper/compScen-ElecH2_Aug22_NewMkup_CESplots-2022-08-18_23.31.56-EU27.pdf``


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> - `Internal|CES Function|Value Added|Input` -> gives value added (price * quantity) of input

That's value, not value added.

chroetz:
> Furthermore, a corresponding section of CES function reporting is added in compare scenarios 2. It is on by default. However, as its output is quite specific one could think about making this section optional for users.

Currently the default sections for compareScenarios2 are in the remind2 package and optional sections are in the REMIND repository. Take a look at https://github.com/remindmodel/remind/pull/923 for an example of such an optional section.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
```
o01_CESmrs <- readGDX(gdx, "o01_CESmrs", restore_zeros = F, react = "silent") # marginal rate of substitution (ratio of CES prices)
```
You are pulling [this calculation](https://github.com/remindmodel/remind/blob/7eeb1038f2cafef38e1224578342206853beff55/modules/01_macro/singleSectorGr/postsolve.gms#L49-L56)
```
*** compute marginal rate of substitution between primary production factors as ratio of CES prices
*** provides the amount of in2 needed to subsitute one unit of in to generate the same economic value
loop((cesOut2cesIn(out,in),cesOut2cesIn2(out,in2))$(ppfen(in)),
  o01_CESmrs(t,regi,in,in2)
  =
  o01_CESderivatives(t,regi,"inco",in) /
  o01_CESderivatives(t,regi,"inco",in2)
  );
  ```
  from the gdx.
  But that (_MRS<sub>1,2</sub> = &part;Y/&part;x<sub>1</sub> / &part;Y/&part;x<sub>2</sub> = p<sub>1</sub>/p<sub>2</sub>_) is true only in equilibrium, i.e. when _p<sub>1</sub> and p<sub>2</sub>_ are unconstrained.  Which means that if any component in the calculation of the derivatives is at a bound (e.g. `ue_steel_primary` due to limits on either the primary steel share or energy efficiency), that is not the "true" MRS anymore.
Not sure it is worth filtering for, but that should be kept in mind when using these figures.

**Comment by fschreyer** [comment]
> Currently the default sections for compareScenarios2 are in the remind2 package and optional sections are in the REMIND repository

Ok thanks. I will leave it in the main section for now. If it turns out to be rarely looked at in standard validations, we still can put it into an optional section later. 

> But that (MRS1,2 = ∂Y/∂x1 / ∂Y/∂x2 = p1/p2) is true only in equilibrium, i.e. when p1 and p2 are unconstrained. [...] that is not the "true" MRS anymore.

I agree that ``CES Price`` is strictly speaking not always the correct naming as it is the derivative of the CES function we are looking at which might not be in equilibrium. However, I personally think that CES price is still a less confusing name for most people and denotes what we usually interpret it as. Surely, it has to be kept in mind that this price can be different from FE price even in absence of mark-up cost due to constraints and that it denotes the marginal product of the production function. 

Concerning the MRS, it is defined as ``∂Y/∂x1 / ∂Y/∂x2``, so as the slope of the isoquants. Not as the ratio of prices. Hence, the MRS is always correctly calculated even if REMIND cannot reach the optimal point where it corresponds to the prices. 


Btw, I put a example PDF above to see what the plots look like.

**Review by 0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q — APPROVED** [review]

File: inst/markdown/compareScenarios2/cs2_10_ces_function.Rmd [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+
+### Buildings
+
+```{r MRS Buildings}
+showLinePlots(data, "Internal|CES Function|MRS|feelhpb|fegab")
+showLinePlots(data, "Internal|CES Function|MRS|feelhpb|fehob")
+showLinePlots(data, "Internal|CES Function|MRS|feelhpb|fesob")
+showLinePlots(data, "Internal|CES Function|MRS|feelhpb|feelrhb")
+
+showLinePlots(data, "Internal|CES Function|MRS|feelh2b|fegab")
```
chroetz, line 108: `Internal|CES Function|MRS|feelh2b|fegab` does not exist. Do you mean `Internal|CES Function|MRS|feh2b|fegab`?

chroetz (changes requested):

inst/markdown/compareScenarios2/cs2_10_ces_function.Rmd
fschreyer: thanks, solved it.

PR #330, opened 2022-08-29
remove function shorthand

pfuehrlich-pik:


fschreyer (approved):
Thanks for the quick fix! :)

PR #334, opened 2022-09-08
change renewable marginal capacity factor calculation

cchrisgong:
change how marginal cost is calculated in reportLCOE, now renewable capacity factor takes the lowest filled grade, not the highest grade that is still free (<90% filled)

The reason we made this change is due to hydro capacity factor in Germany in REMIND and in reporting are not aligned. 

Robert:
"Now for hydro, it seems REMIND stays always inside grade 4, but fills it (as much as possible with the deltacap/2-zigzag), so that the model sees the CF of grade 4, while the reporting sees the CF of grade 5
I would change the reporting and use the last grade that is still filled
this will then slightly overestimate CF for the cases that the model expands this technology, but in such a case it will anyway go into the next bad grade in the next time step, so it should in general look better"

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '212008170'
+ValidationKey: '212250290'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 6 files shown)


robertpietzcker (approved):
Seems good to me

PR #336, opened 2022-09-20
increase reporting compatibility with older runs

Renato-Rodrigues:
Extending the solution introduced in this commit https://github.com/pik-piam/remind2/commit/b59d0a3b0fc2f301783aeec36679a895c6fab1ec to increase compatibility with older runs like: `/p/projects/remind/users/renatoro/ECEMF/v04_2022_04_06/output/04_DIAG-NZero_2022-04-08_16.53.53`

### DIFF
--- a/man/readSupplycurveBio.Rd
+++ b/man/readSupplycurveBio.Rd
@@ -6,7 +6,9 @@
 \usage{
 readSupplycurveBio(
   outputdirs,
-  userfun = function(param, x) {     return(param[[1]] + param[[2]] * x) },
+  userfun = function(param, x) {
+     return(param[[1]] + param[[2]] * x)
+ },
   mult_on = "all"
 )
 }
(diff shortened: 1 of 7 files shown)


man/readSupplycurveBio.Rd
```diff
@@ -6,7 +6,9 @@
 \usage{
 readSupplycurveBio(
   outputdirs,
-  userfun = function(param, x) {     return(param[[1]] + param[[2]] * x) },
+  userfun = function(param, x) {
```
fschreyer, line 9: Is that intended?

fschreyer (approved):

man/readSupplycurveBio.Rd
robertpietzcker: it is just two line breaks added, right?

Renato-Rodrigues: The Rd changes (as well as git related ones) are created automatically by the lucode2 buildLibrary code.
I asked RSE to remove the inclusion of these changes in content related PRs. 
However, after discussing this in the mattermost group, there was no common agreement at the time.

PR #338, opened 2022-09-27
Make getMifScenPath find reportings also from coupled runs

dklein-pik:
Remove old function that was making up the reporting filename from the name of the results folder and was relying on result folders that end with a time stamp. Now: explicitly searching for existing mif files.

### DIFF
--- a/R/getPath.R
+++ b/R/getPath.R
@@ -1,8 +1,3 @@
-getScenNamesFast <- function(outputDirs) {
-  folder <- basename(outputDirs)
-  substr(folder, start = 0, stop = nchar(folder) - 20)
-}
-
 #' Get Paths to Certain Files in the REMIND Directory
 #'
 #' \code{getMifScenPath}: get path to the scenarios' reporting mifs.
@@ -16,8 +11,10 @@ getScenNamesFast <- function(outputDirs) {
 #' @rdname getPath
 #' @export
 getMifScenPath <- function(outputDirs, mustWork = FALSE) {
-  names <- getScenNamesFast(outputDirs)
-  path <- file.path(outputDirs, paste0("REMIND_generic_", names, ".mif"))
+  # Find all mif files starting with REMIND_generic_. The result will also contain REMIND_generic_withoutPlus.mif
+  reports <- list.files(outputDirs, "^REMIND_generic_.*\\.mif$", full.names = TRUE)
+  # Filter out filenames having "withoutPlus.mif" in their name
+  path <- reports[grepl("REMIND_generic_(?!.*(withoutPlus)\\.mif).*\\.mif$", reports, perl = TRUE)]
   normalizePath(path, mustWork = mustWork)
 }
 
--- a/R/getPath.R
+++ b/R/getPath.R
@@ -11,11 +11,9 @@
 #' @rdname getPath
 #' @export
 getMifScenPath <- function(outputDirs, mustWork = FALSE) {
-  # Find all mif files starting with REMIND_generic_. The result will also contain REMIND_generic_withoutPlus.mif
-  reports <- list.files(outputDirs, "^REMIND_generic_.*\\.mif$", full.names = TRUE)
-  # Filter out filenames having "withoutPlus.mif" in their name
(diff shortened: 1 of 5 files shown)


R/getPath.R
```diff
@@ -16,8 +11,10 @@ getScenNamesFast <- function(outputDirs) {
 #' @rdname getPath
 #' @export
 getMifScenPath <- function(outputDirs, mustWork = FALSE) {
-  names <- getScenNamesFast(outputDirs)
-  path <- file.path(outputDirs, paste0("REMIND_generic_", names, ".mif"))
+  # Find all mif files starting with REMIND_generic_. The result will also contain REMIND_generic_withoutPlus.mif
+  reports <- list.files(outputDirs, "^REMIND_generic_.*\\.mif$", full.names = TRUE)
+  # Filter out filenames having "withoutPlus.mif" in their name
+  path <- reports[grepl("REMIND_generic_(?!.*(withoutPlus)\\.mif).*\\.mif$", reports, perl = TRUE)]
```
mikapfl, line 17:
```suggestion
  path <- reports[grepl("^REMIND_generic_(?!.*(withoutPlus)\\.mif).*\\.mif$", reports, perl = TRUE)]
```
Or is there a reason to avoid `^` here? For me, it is always easier to think about regexen with `^` and `$`, because I don't have to think about the effects of the implicit `.*` at the start or end, respectively.

mikapfl: Also, couldn't there be multiple mifs? Then, `path` might be a vector with length > 1, and everything explodes?

mikapfl: Hmmm… if you do it in two steps, you can also radically simplify the `grep`, I think:

```R
path <- grep("withoutPlus\\.mif$", reports, invert = TRUE)
```
does the job, right?

mikapfl: Argh, you also need a `value = TRUE`:
```R
path <- grep("withoutPlus\\.mif$", reports, invert = TRUE, value = TRUE)
```

R/getPath.R
```diff
@@ -16,8 +11,10 @@ getScenNamesFast <- function(outputDirs) {
...
   normalizePath(path, mustWork = mustWork)
```
mikapfl, line 18:
```suggestion
  return(normalizePath(path, mustWork = mustWork))
```
I know you didn't change it, but I think an explicit return would make this much easier to read.

R/getPath.R
orichters: Note that further .mif files could be present, such as `_adjustedPolicyCosts.mif`. I think it would be safer to use https://github.com/pik-piam/lucode2/blob/master/R/getScenNames.R

dklein-pik: > Or is there a reason to avoid `^` here? For me, it is always easier to think about regexen with `^` and `$`, because I don't have to think about the effects of the implicit `.*` at the start or end, respectively.

Yes. Since `list.files(..., full.names = TRUE)` returns the full paths (not only the file name), the string does not start with "REMIND...".

dklein-pik: > Hmmm… if you do it in two steps, you can also radically simplify the `grep`, I think:
> 
> ```r
> path <- grep("withoutPlus\\.mif$", reports, invert = TRUE)
> ```
> 
> does the job, right?

True. Thanks. Btw: didn't know the `invert` argument :-)

dklein-pik: > Note that further .mif files could be present, such as `_adjustedPolicyCosts.mif`. I think it would be safer to use https://github.com/pik-piam/lucode2/blob/master/R/getScenNames.R

Ok, if this is the case, we should indeed use getScenNames.R I wanted to avoid this because to me it seems a bit more time-consuming, but maybe not relevant here.

orichters (approved):
Thanks! I also tend to make backups of mif files before redoing the reporting, so this is the safe solution. Indeed, the old function is quicker, but the new one is still well below a tenth of a second, so I think that is ok :)

mikapfl (commented):
Nice, thanks!

PR #344, opened 2022-10-25
Add piamInterfaces intergration for AR6 and NGFS

fbenke-pik:
This PR extends the testing of remind2 by adding a check making sure that all the REMIND variables currently used in AR6 and NGFS mapping templates are generated by the remind2 reporting. 

**Details**

In this poc, i download an NGFS gdx, run the reporting gdx2mif in remind2 and then compare the variables in the generated mif file with the ones expected in the NGFS project template. This could be a blueprint for writing tests for all the other projects, i.e. SHAPE, NAVIGATE, ARIADNE ...

However, there are two disadvantages to this approach:

1. These tests for now can only be executed on the local machine as there is no support of the gdx api for the github workflow. So if we want to make sure these tests are run before a PR can be merged, the contributor must have GAMS and the gdx api installed on their local machine, otherwise the aforementioned test will fail and buildLibrary does not finish.  
2. If we keep adding a test for each project (which seems inevitable, as one gdx for all projects wont work due to project-specific variables), the time to run all the checks on your local machine could increase significantly. We try to cushion this right now by running tests in parallel. Using oneRegi gdx files does not seem to reduce runtime.

If these tests become too much of a burden for the remind2 development, we can turn them off by default and add a TODO in the documentation to run them manually only in case of reporting adjustments.


tests/testthat/test-convGDX2mif.R
```diff
@@ -92,10 +94,13 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
   # gdxPaths <- c(gdxPaths, .findAMTgdx())
 
   numberOfMifs <- 0
+  computedVariables <- NULL
+
   for (gdxPath in gdxPaths) {
     numberOfMifs <- numberOfMifs + 1
     message("Running convGDX2MIF(", gdxPath, ")...")
     mifContent <- convGDX2MIF(gdxPath, gdx_ref = gdxPath)
+    computedVariables <- unique(c(computedVariables, getItems(mifContent, dim = 3.3)))
```
fbenke-pik, line 103: As a first step, this will check all the mif files generated from gdx files for available REMIND variables in remind2 reporting. 

We might have to add additional gdx files, if the existing ones don't cover all the configurations to be run when running reporting for specific projects.

fbenke-pik: Also, this is not a specific testing where a project gdx is compared to the REMIND variables of the corresponding project template (not sure if this is necessary).

orichters: I think that is a great start. Still thinking whether it may be worth adding a parameter  to `getREMINDTemplateVariables()` such as `type = standalone` or `REMIND-MAgPIE` or however this should be called.

fbenke-pik: Don't know enough about coupled runs and reporting to understand what difference it makes? 

Would we need two tests, one for standalone run with a standalone gdx, and one for a REMIND-MAgPIE with a gdx representing a coupled run?

orichters: Coupled runs simply have much more variables related to land-use, agriculture etc. that they get from MAgPIE. But they are not part of the REMIND gdx file, but are attached to the mif file later, [in reporting.R](https://github.com/remindmodel/remind/blob/3b719d8dd4856b709a793cb34b9e0b35082244ff/scripts/output/single/reporting.R#L86-L103). Therefore, I don't think a remind2 test can cover these, but rather the MAgPIE people should set up tests for that, I guess.

PR #345, opened 2022-11-04
add switchValuesScenConf()

orichters:
- if you call it with `switchValuesScenConf("cm_startyear")` in the remind directory as an example, it prints what values are used in different scenario config files and main.gms and returns a vector of these values.
- if you call it as `switchlist <- switchValuesScenConf()`, it prints and returns all the switches defined in main.gms that in all scenario config files have one single value that you can access with `switchlist["cm_wind_offshore"]` to get the value (you can also look in main.gms, of course).


Renato-Rodrigues (approved):
Thks!!

PR #348, opened 2022-11-14
allow p21_CO2TaxSectorMarkup to have time-dependence

orichters:
- as preparation for a remindmodel PR, allow p21_CO2TaxSectorMarkup to have time-dependence, without breaking the reporting if it doesn't


christophbertram (approved):
I don't really understand what this does but if it does what is said, then it's useful...

orichters:
I cleaned it a bit, maybe it is more understandable now.

PR #351, opened 2022-11-16
Fix/remindmodel/remind/1088

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
In case you do actually need this.


orichters (approved):
Thanks for implementing it. I don't think I will need it, I already started to rerun the scenario yesterday. Still, once you already did it, may be worth keeping it, I leave that up to you.

PR #353, opened 2022-11-21
Bugfix: Split of sectoral solids

robinhasse:
The split of industry and buildings solids into traditional and modern biomass as well as coal ignored that solids have been split into biomass and fossil before already. This caused inconsistent quantities. I corrected the biomass split into modern and traditional such that it sums up to biomass. All fossil solids are considered to be coal. 

Moreover, I introduced a `|+|` to the biomass level as well. I am unsure if this renaming might cause problems elsewhere e.g. in exogains? Would you know, @Renato-Rodrigues?

grafik

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '2130383968'
+ValidationKey: '2130974172'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.103.14",
+  "version": "1.103.16",
   "description": "<p>Contains the REMIND-specific routines for data and model\n    output manipulation.<\/p>",
   "creators": [
     {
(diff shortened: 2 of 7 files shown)


orichters:
What needs update is probably the [mapping templates](https://github.com/pik-piam/piamInterfaces/tree/master/inst/templates)…

robinhasse:
> Great you fixed this bug. What needs update is probably the [mapping templates](https://github.com/pik-piam/piamInterfaces/tree/master/inst/templates)…

Thanks @Renato-Rodrigues, I renamed the variable [here](https://github.com/pik-piam/piamInterfaces/pull/37).

orichters:
I wasn't Renato, but I would check `reposearch -F "FE|Industry|Solids|Biomass|"` on the cluster to see where possible adaptations have to made because of the introduction of the `+`. exoGains seems [to delete the `+` anyway](https://github.com/orichters/remind/blob/develop/scripts/input/exoGAINSAirpollutants.R#L48), but remind2::reportFE should certainly be checked.

robinhasse:
> I wasn't Renato, but I would check `reposearch -F "FE|Industry|Solids|Biomass|"` on the cluster to see where possible adaptations have to made because of the introduction of the `+`. exoGains seems [to delete the `+` anyway](https://github.com/orichters/remind/blob/develop/scripts/input/exoGAINSAirpollutants.R#L48), but remind2::reportFE should certainly be checked.

Thanks for the hint to `reposearch`, I was not aware. I already found that the `|+|` is irrelevant for exogains. And `remind::reportFE` is the file I change in this PR, or do you mean something else, @orichters?

orichters (approved):
looks convincing, thanks!

PR #354, opened 2022-11-28
in cs2 fix y-axis for area plots and write warnings to log, not PDF. + Add all contributors

orichters:
- I was always a bit confused by cs2 plots as the automatic adaptation of the y-axis often made it difficult to compare scenarios. This PR together with https://github.com/pik-piam/mip/pull/48 fixes the y-axes for the area plots, making the differences between the scenarios more obvious, see the differences below
- full files available at `/p/tmp/oliverr/NGFS_v3_tests/2022-11-16/remind` (`compScen-47tests-2022-11-23_16.51.17-H12.pdf` vs. `compScen-47tests-2022-11-23_16.51.17-H12-new.pdf`)
- for most variables, it simply improves readability and comparability
- for CDR that is often close to zero in current policy runs, this means you can hardly see anything, but I think this is rather a boon as it reflects the scenario reality, than a drawback, but maybe @katarkow or @strefler can comment whether I undo the CDR plots
- add all contributors to the author list as agreed with @Renato-Rodrigues a while ago

left: old, right: new

- cumulative CO2
test-cs2-fixed-yaxes-co2

- Capacity Electricity
test-cs2-fixed-yaxes-cap-elec

- FE CDR
test-cs2-fixed-yaxes-cdr

- Emi CDR
test-cs2-fixed-yaxes-cdr-2

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '2132005668'
+ValidationKey: '2132135313'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,6 +1,6 @@
 {
   "title": "remind2: The REMIND R package (2nd generation)",
-  "version": "1.103.18",
+  "version": "1.103.19",
   "description": "<p>Contains the REMIND-specific routines for data and model\n    output manipulation.<\/p>",
   "creators": [
     {
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -5,6 +5,90 @@
   "creators": [
     {
       "name": "Rodrigues, Renato"
+    },
+    {
+      "name": "Baumstark, Lavinia"
+    },
+    {
+      "name": "Benke, Falk"
+    },
+    {
+      "name": "Dietrich, Jan Philipp"
+    },
+    {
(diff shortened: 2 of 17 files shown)


Renato-Rodrigues (approved):

orichters:
> Regarding the section headlines that you also added again - are you aware of the reasons why Christof removed them [a while ago](https://github.com/pik-piam/remind2/pull/302/files#r924579931)? In his opinion:
> 
> > ...they are redundant and if you happen to label two sections with the same name, everything crashes. This can easily happen if you split a chunk via shortcut; then both resulting chunks have the same name. So better to not use labels for chunks unless they are actually useful.

Thanks, Pascal, for this warning. I actually find the log much more helpful with them instead of just reading `label: unnamed-chunk-4` all the time.

orichters:
We may use `options(knitr.duplicate.label = "allow")` as suggested [here](https://bookdown.org/yihui/rmarkdown-cookbook/duplicate-label.html)

orichters:
I did as suggested. All those annoying warnings in the compareScenarios2 pdf now are moved to the log file, reducing the file size by 58 pages. In this process, I had some problems with tests failing, and one reason was that prices can become infinite if the capital factor is zero, particularly, `Price|Final Energy|Transport|Gases|Natural Gas|Total LCOE` and `Price|Final Energy|Transport|Gases|Natural Gas|Transport and Distribution` and their moving averages. I fixed that now: If the capacity factor is 0, price is also set to zero.

PR #356, opened 2022-12-02
don't let buildLibrary fail if variable name incompatible with piamInterfaces

orichters:
- will still raise a warning
- fails only if no variable name matches (then, probably something went deeply wrong)


fbenke-pik (approved):
Yes, i am fine with this solution for now. Especially, since once you updated piamInterfaces, you must immediately update remind2 as well, otherwise other open remind2 PRs will not build. 

Note that in the long run, we might disable these tests per default.

PR #362, opened 2022-12-14
bugfix LCOE CO2 price to incorporate regional CO2 price in regional ETS

cchrisgong:


R/reportLCOE.R
```diff
@@ -307,7 +308,7 @@ reportLCOE <- function(gdx, output.type = "both"){
                   getNames(v_emiTeDetail[,,"co2"], dim = 3))
 
  te_annual_co2_cost[,,tmp] <- setNames(
-   ( pm_priceCO2[,getYears(v_emiTeDetail),]
+   ( (pm_priceCO2[,getYears(v_emiTeDetail),] + pm_taxemiMkt[,getYears(v_emiTeDetail),"ETS"])
    * dimSums(v_emiTeDetail[,,tmp], dim = c(3.1, 3.2, 3.4), na.rm = TRUE)
    * 1e9   # $/tC * GtC/yr * 1e9 t/Gt = $/yr
    ),
```
Renato-Rodrigues, line 314: I am not that familiar with the LCOE code, but your code is using the ETS price (`pm_taxemiMkt[,,"ETS"]`) and applying it to all possible emissions (`v_emiTeDetail`).

I would instead define this part of the code inside a if else statement with two parts. 
1. The first one would contain code that would only be applied to regions that do not have `pm_taxemiMkt` prices.
 - the code inside this if statement would correspond exactly to the original code.
2. The second part of the if statement (the else part) would be applied to regions that have values in `pm_taxemiMkt`.
- inside of this code you would need to have three different `te_annual_co2_cost` components. One for the ETS, one for the ESR, and one for the other emission market calculation. 
- you would use `pm_taxemiMkt` instead of `pm_priceCO2` and `vm_emiTeDetailMkt` and instead of `v_emiTeDetail` in the calculation.
- only then you would sum up these three components to get the corresponding `te_annual_co2_cost` for the tech.

Renato-Rodrigues (changes requested):

R/reportLCOE.R
```diff
@@ -124,7 +125,7 @@ reportLCOE <- function(gdx, output.type = "both"){
  v_investcost  <- readGDX(gdx,name=c("vm_costTeCapital","v_costTeCapital","v_investcost"),field="l",format="first_found")[,ttot,]
  vm_cap        <- readGDX(gdx,name=c("vm_cap"),field="l",format="first_found")
  vm_prodFe     <- readGDX(gdx,name=c("vm_prodFe"),field="l",restore_zeros=FALSE,format="first_found")
- v_emiTeDetail <- readGDX(gdx,name=c("vm_emiTeDetail","v_emiTeDetail"),field="l",restore_zeros=FALSE,format="first_found")
+ v_emiTeDetail <- readGDX(gdx,name=c("vm_emiTeDetail","v_emiTeDetailMkt"),field="l",restore_zeros=FALSE,format="first_found")
```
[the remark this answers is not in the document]
cchrisgong: addressed

R/reportLCOE.R
cchrisgong: Thank you for the detailed suggestions! I made the new implementation (see below)

The only question I have is:
I summed over the gas-type dimension in vm_emiTeDetailMkt by doing
"dimSums(vm_emiTeDetailMkt[reg,,tmp], dim = c(3.1, 3.2, 3.4, 3.5), na.rm = TRUE)"

this will essentially sum over the co2 price, and no2 price, etc together to give a "CO2 price", is this correct? The non-CO2 price usually is a small component of the whole, so it probably doesn't matter that much but I want to check with you

cchrisgong: also what is the unit of the pm_taxemiMkt? my guess is T$/GtC but not sure (maybe it can be added to declaration.gms)

Renato-Rodrigues (approved):

PR #364, opened 2022-12-16
allow snake_case

pfuehrlich-pik:


orichters (approved):
thanks! We should have a discussion about line length etc. as well, either we put in some effort to format it correctly, or we should decide not to do it (or go to 150 or so).

pfuehrlich-pik:
Sure, we can discuss :wink: I'm strongly in favor of keeping the 120 characters limit, and re-format the code accordingly, because in my opinion it's not hard to split up lines. Also, even in cases where for some reason it's really hard to get below 120 because of e.g. excessively long variable names and/or where the code is not really meant to be read (e.g. data is hardcoded) I'd much rather use `# nolint` than disabling this linter rule altogether.
Increasing the limit to 150 is the worst of both worlds - some lines are still too long, and all lines approaching that limit do not fit on a screen (120 is already too much for smaller laptops), so I'd rather disable the rule altogether before increasing to 150.

mikapfl (approved):
Looks good in general, but I haven't worked in remind2 so far, so I can't really comment on the reasons for overly long lines here.

orichters:
@mikapfl: Basically, we have [many of lines similar to those and the following](https://github.com/orichters/remind2/blob/master/R/reportFE.R#L133). I do not have a strong preference whether that is a good idea, I mainly want to have lintr warnings that are meaningful and reflect the actual state of the package.

I suggest to have a discussion between say @pfuehrlich-pik and @Renato-Rodrigues which I'm willing to join to bring in the user perspective or as a moderator.

pfuehrlich-pik:
I guess my first question would be, is this nicely readable code that should just get a thumbs up by the linter? I think that's quite hard to argue - but then the next question of course would be, is that important in this case? Maybe not, and in that case I'd say just `# nolint` the entire function, that's still much better than disabling the rule for the entire existing package and all code that will be written in the future in remind2.

pfuehrlich-pik:
To nolint the entire function we don't need `# nolint` in every line, instead we could do
```
# nolint start: line_length_linter.
...
# nolint end
```

pfuehrlich-pik:
I the case of `reportFE` I don't think this radical option is not necessary though, I'm quite confident it could be refactored to stay within 120 characters per line in less than 3 hours I'd estimate.

Renato-Rodrigues:
The code could be refactored indeed, but I see no reason spending time with older code just to abide with linter rules, while there are so many other priorities in the table. 

Besides, there is no reason to believe that the rewrote code would be better in any way. Following a column limitation set due to hardware constraints from 40 years ago does not necessarily help anything. Whoever develops this library using a terminal with fixed number of column items should be presented to modern ways of coding in the first place.   

Don't get me wrong, formatting rules can be definitely a good thing and I applaud the effort. I would have a completely different opinion if we were starting a new project, but we are working with a lot of baggage here.

I would say that a much better approach on libraries with a lot of extra baggage would be to issue the linter warnings only for newly added or changed code lines. This way things could improve gradually.

The way it is now, it is all or nothing. There is very little incentive to whoever is committing a change to even care about the linter warnings.
Simply disabling them for the entire function or including them on the majority of lines of a function would also serve no purpose in my opinion to improve the situation.

pfuehrlich-pik:
> Besides, there is no reason to believe that the rewrote code would be better in any way.

I disagree. Reading, changing and adding to the code would be much faster and could be done with more confidence of not introducing bugs by more people (not just the people who originally wrote the code).

This is not about hardware, even if there was a screen that could display 500 characters in one line or you are using an editor with line wrap this is still a problem. It's about human mental capabilities. If too much happens on a line a human cannot comprehend it which is problematic when reading the code, but also when writing it. Lines with more characters beyond some threshold are exponentially more likely to have bugs in them.

> issue the linter warnings only for newly added or changed code lines

Completely agree. Unfortunately the linter we are using has the technical limitation of working on a per file basis. However, by default `lucode2::lint` or `lucode2::buildLibrary` will only lint files that were modified since the last version number increase, so it is only all or nothing for a file you have touched. Admittedly for functions with 1500+ lines of code this quite the opposite of gradual, but in general I'd say it's close enough.

> Simply disabling them for the entire function or including them on the majority of lines of a function would also serve no purpose in my opinion to improve the situation.

Why not? This way old code does not need to be rewritten/spend time on, but new code will be checked.

Renato-Rodrigues:
With the risk of hijacking even more this PR, my main points are:

 (1) we should avoid wasting resources on unnecessary code rewrite. There is a huge number of priorities in line before messing up with things that are already tested and working.
 (2) we should only provide reasonable warnings instead of overflowing developers with either a huge number of irrelevant information - which causes people to simply ignore all linter warnings together - or not provide any warnings at all - by disabling the linter checks on entire functions or cluttering the file with a lot of disable lines.

I don't think there is any discussion here about the mentioned code lacking human readability and/or being complicated to understand.
The same code was also understood and modified innumerous times by a high number of people with different backgrounds.
Therefore, the hypothetical mental benefit of limiting the number of line characters is not a given per se in the discussed case.

If we accept these facts, as this code recipe is used in the majority of files of the remind2 package, you are going to either need to lose a lot of time rewriting code, or you are going to disable linter checks in a huge amount of cases (either files or multiple lines).
Besides, the existent code, although certainly not being bug free, was already put to test innumerous times. To guarantee that you achieve the same level of reliability after you rewrite anything, you need to dedicate a substantial amount of time which is detrimental in the available time to invest in other issues that actually improve the remind model.

PR #365, opened 2022-12-21
unify capacity reporting

fbenke-pik:


orichters (approved):
thanks for catching the inconsistent naming and the wrong units!

PR #366, opened 2022-12-29
debug power demand in reportSE

cchrisgong:


Renato-Rodrigues (changes requested):
I don't agree with the changes in the reportSE function. 
If you control the variable for getting only the effort sharing part of the final energy deamnd, you are simply ignoring the activities of the sector that belong to the ETS. This is specially problematic for the industry sector.

cchrisgong:
Do you have a suggestion how I can modify this? Should I sum over ETS instead of ES then? (or combining them?)

cchrisgong:
Some preliminary plots show summing by ES seems fine (for showing breakdown of power demand, negative axis): 
image

Renato-Rodrigues:
the variable `SE|Input|Electricity|Industry (EJ/yr)` does not have the emission market dimension, therefore it should sum up for all emission markets (ETS, ES, other,...).
The original code made sure that happens. Your change breaks that.

I see no reason why the original code would give any issue. If you want the effort sharing only quantities, then you need to create a new variable with the effort sharing dimension, e.g. `SE|Input|Electricity|Industry|ESR (EJ/yr)`

cchrisgong:
sorry wasn't being clear before. . The original code was broken, and doesn't sum it over. It results in SE|Input|Electricity|Industry (EJ/yr) giving NAs for variables after start_year (below from the newest AMT run: /p/projects/remind/modeltests/output/SSP2EU-AMT-PkBudg670_2022-12-31_14.14.47).
image

 I looked and found that it is likely because vm_demFeSector used to NOT have an emission market dimension, but now it does. So I have to add a dimension so the script produces some number.

Yes, I would like to have an emission market independent (or summed over) value for SE|Input|Electricity|Industry (EJ/yr). How do I best write this in R when vm_demFeSector has now an extra dimension?

pweigmann:
> sorry wasn't being clear before. . The original code was broken, and doesn't sum it over. It results in SE|Input|Electricity|Industry (EJ/yr) giving NAs for variables after start_year (below from the newest AMT run: /p/projects/remind/modeltests/output/SSP2EU-AMT-PkBudg670_2022-12-31_14.14.47). image
> 
> I looked and found that it is likely because vm_demFeSector used to NOT have an emission market dimension, but now it does. So I have to add a dimension so the script produces some number.
> 
> Yes, I would like to have an emission market independent (or summed over) value for SE|Input|Electricity|Industry (EJ/yr). How do I best write this in R when vm_demFeSector has now an extra dimension?
> 
> Thanks!!

I think, this can be solved by simply replacing the NAs in vm_demFeSector with zeros via
`vm_demFeSector[is.na(vm_demFeSector)] <- 0`
maybe right after the definition of the object in line 370. I currently have problems to build the library, so I can't commit the changes myself today.

cchrisgong:
Thanks, @pweigmann for looking into it. However, it is not a matter of NA replaced with 0, because the variables are supposed to have non-zero values - these are demand variables. The model certainly is not proposing us have zero demand after 2020. One must improve the way this is summed up with the extra dimension of emission market. I hope I made it clear now.. :D thanks a lot

cchrisgong:
Ahh thanks @pweigmann for the clarification. I see what you meant! If you are right, this could fix it (and quite easily too) :)

pweigmann:
Hey Chris, it was easier for me to just create a new PR with my change, hope that's ok for you! See #373

cchrisgong:
perfetto! danke!

Renato-Rodrigues:
Thank you for both of you to work on solving this!!

PR #373, opened 2023-01-11
bugfix FE demand in reportSE

pweigmann:
as discussed in  #366


Renato-Rodrigues (approved):
This change should only affect in positive ways any further calculations in the function.
As long as you checked a mif file and saw that output numbers made sense to fix the initial issue you had, I would consider the issue solved.
Thank you!

PR #379, opened 2023-01-25
Add and rename floor space variables.

robinhasse:
Adds extra floor space variables for residential and commercial Buildings if present in the gdx. Also, I rename these variables from `Energy Service|*` to `ES|*` to be consistent with transport variable names. Variable mappings in piamInterfaces are changed in [#76](https://github.com/pik-piam/piamInterfaces/pull/76).

### DIFF
--- a/R/reportFE.R
+++ b/R/reportFE.R
@@ -20,7 +20,7 @@
 #' @export
 #' @importFrom gdx readGDX
 #' @importFrom magclass new.magpie mselect getRegions getYears mbind setNames
-#'                      dimSums getNames<- as.data.frame as.magpie
+#'                      dimSums getNames<- as.data.frame as.magpie getSets
 #' @importFrom dplyr filter %>% mutate select group_by summarise left_join full_join
 #'                   ungroup rename
 #' @importFrom quitte inline.data.frame revalue.levels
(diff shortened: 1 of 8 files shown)


R/reportFE.R
```diff
@@ -539,7 +539,15 @@ reportFE <- function(gdx, regionSubsetList = NULL,
 
   p36_floorspace <- readGDX(gdx, "p36_floorspace", react = "silent")[, t, ]
   if (!is.null(p36_floorspace)) {
-    out <- mbind(out, setNames(p36_floorspace, "Energy Service|Buildings|Floor Space (bn m2/yr)"))
```
orichters, line 542: Do I understand that correctly that `p36_floorspace ` is either one single component (`buildings`) or three, and the code is flexible enough to handle both situations?

R/reportMacroEconomy.R
```diff
@@ -160,9 +160,6 @@ reportMacroEconomy <- function(gdx, regionSubsetList = NULL,
                                  + invM[, , "Investments|Non-ESM|Macro (billion US$2005/yr)"],
                                  "Investments|Non-ESM (billion US$2005/yr)"))
 
-    # add floorspace
-    invM <- mbind(invM, setNames(p36_floorspace[, getYears(invM), ] * 1000, "Floorspace demand (million m2)"))
```
orichters, line 164: old legacy that never made it into piamInterfaces?

orichters (approved):

R/reportFE.R
robinhasse: Yes, the parameter is just passed through to get here. See [#1197](https://github.com/remindmodel/remind/pull/1197)

R/reportMacroEconomy.R
robinhasse: Probably.

PR #380, opened 2023-01-27
Bugfix for residential & commercial floor space.

robinhasse:
This corrects a bug of #379 that had no effect for old remind runs but let new runs with more detailed floor space data that it was introduced for fail.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '214403684'
+ValidationKey: '214445192'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


orichters (approved):
don't really understand what the problem is, but I think you know what you are doing :)

robinhasse:
> don't really understand what the problem is, but I think you know what you are doing :)

Well, I hope so, too. the test for the existence of a 3rd dimension in the MAgPIE object was wrong and therefore never fulfilled.

PR #382, opened 2023-02-13
adding investment cost with adjustment cost to LCOE reporting

cchrisgong:


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
No idea what this code does, as there is no documentation, but it does not seem to crash.
¯\_(ツ)_/¯

cchrisgong:
Felix and I are the only ones using this script (for now), it might be good to add some documentation at some point though

PR #383, opened 2023-02-14
Revert "adding investment cost with adjustment cost to LCOE reporting"

cchrisgong:
Reverts pik-piam/remind2#382


cchrisgong:
Forgot to build library...

PR #385, opened 2023-02-17
Add FE price index to reporting

merfort:


orichters:
@merfort: Seeing this by chance now.

1. May it be preferable to use the structure `Prices|Index2010|Agriculture|Food products|Plant-based|Maize` as the MAgPIE people do, see [here](https://github.com/pik-piam/piamInterfaces/blob/master/inst/templates/mapping_template_AR6.csv#L166-L172)? Apart from that note, that [`piamInterfaces::priceIndicesIIASA()`](https://github.com/pik-piam/piamInterfaces/blob/master/R/priceIndicesIIASA.R) calculates the price indices automatically, if they are provided in the `iiasatemplate`. I don't know if you need these values earlier in remind2 format, by we may want to harmonize our work at some point :)

fbenke-pik:
Closing, due to no activity.

PR #388, opened 2023-03-13
adapt colorScenConf, compareScenConf to new remind start scripts

orichters:
- bugfix


mikapfl (approved):
Thanks, I had that on a TODO list somewhere still.

orichters:
> Thanks, I had that on a TODO list somewhere still.

Would all be easier if remindmodel was an `R` library :)

PR #389, opened 2023-03-16
Report net GDP after damages and fix aggregation of damage factors

gabriel-abrahao:
Adds the calculation of net GDP, that is, GDP after damages, to the reporting. And fixes a bug in which the damage factors were aggregated incorrectly

The choice of name for the GDP variable may be contentious, please take a look @LaviniaBaumstark @fpiontek . What we currently report in `GDP|MER` (and `GDP|PPP`) is the gross GDP. In runs with damages set up, the net GDP is supposed to be that times `Damage factor`. I don't know if we want to change the name of such an important variable, so currently I just named the net GDP ones `GDP|MERNet_afterDamages` and `GDP|PPP|Net_afterDamages`. @fpiontek please see if you'd prefer a different name.

It also fixes the aggregation of the damage factor. Before, it was just aggregate in a sum with all the other variables. Now it's first aggregated as the mean between regions weighted by `GDP|MER`, and then added to `out`.

### DIFF
--- a/R/reportMacroEconomy.R
+++ b/R/reportMacroEconomy.R
@@ -342,7 +342,7 @@ reportMacroEconomy <- function(gdx, regionSubsetList = NULL,
 
 
   # define list of variables that will be exported:
-  varlist <- list(cons, gdp, gdp_ppp, invE, invM, pop, cap, inv, ces, damageFactor, welf) # ,curracc)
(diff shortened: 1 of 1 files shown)


fpiontek:
I am fine with the name and would agree to leave the "old" GDP variable untouched.

fpiontek (approved):

LaviniaBaumstark:
I agree, keep the "old" name

LaviniaBaumstark (approved):

PR #400, opened 2023-04-20
cs2: added a plot for LDV stock and sales

joanna-sitarz:
cs2 was using outdated variable names for LDV sales and stock and thus no plots were generated
I included 2 new plots with variable names matching REMIND-EU .mif names


fbenke-pik (approved):
Looks good to me.

PR #401, opened 2023-04-20
add Price|Carbon|Supply to regipol reporting

orichters:
`Price|Carbon|Supply` was not updated if regipol was used, leading to ~ 0 prices in EUR NDC SSP2EU.

I just give it the same value as `Price|Carbon`.

image

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '215569497'
+ValidationKey: '215655426'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'remind2: The REMIND R package (2nd generation)'
-version: 1.107.7
-date-released: '2023-04-14'
+version: 1.107.8
+date-released: '2023-04-20'
 abstract: Contains the REMIND-specific routines for data and model output manipulation.
 authors:
 - family-names: Rodrigues
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: remind2
 Title: The REMIND R package (2nd generation)
-Version: 1.107.7
-Date: 2023-04-14
+Version: 1.107.8
+Date: 2023-04-20
 Authors@R: c(
     person("Renato", "Rodrigues", , "renato.rodrigues@pik-potsdam.de", role = c("aut", "cre")),
     person("Lavinia", "Baumstark", role = "aut"),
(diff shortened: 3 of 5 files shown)


robertpietzcker:
Hi Renato, 
I don't yet fully understand the handling of the prices between split emiMarkets and a uniform, so maybe my comment misses the point. 

Just from what you write above, I think we won't expect a case where the regions for which emiMarket is turned or off change - the EU is the only place where we have a good reason for the split of emission markets, as we don't go into such detail for any other region. 

on the other point
> the conditionals if(!is.null(pm_taxemiMkt)) and if(!is.null(p47_taxCO2eq_AggFE)) will be always true if t includes also the fixed reference years.

is this a different `t` than the `t` in REMIND? if it is the same, then it will not contain the years for which the run was fixed, as fixing is defined by "not including t"?

Renato-Rodrigues:
`t` in this case refers to the remins2 reporting parameter in the reportign functions, e.g. reportPrices, and by default include 2005 as we almost always report all year variables

robertpietzcker:
ok, thanks for the explanation

Renato-Rodrigues:
> Just from what you write above, I think we won't expect a case where the regions for which emiMarket is turned or off change - the EU is the only place where we have a good reason for the split of emission markets, as we don't go into such detail for any other region.

That is not true. Our default scenario config `scenario_config.csv` has `cm_emiMktTarget` values for NPI and NDC, and it is `off` for all ambition mitigation cases. I did not checked validation results, but there is potentially quite some of price variables wrongly reported because of this issue.

Renato-Rodrigues:
The main point is, our prices reporting is much more complicated then it needs to be because of the need to handle these special cases simply because we use two parameters for the same thing in the model: `pm_taxemiMkt` and `pm_taxCO2eq `

robertpietzcker:
sorry, I typed too fast. I didn't want to include the cases where the emiMarket is turned off - this is exactly what caused Oliver to find the problem, and then to write this patch to overcome it. I understood that you worry about individual regions being included or taken out of the emiMarket , without it being completely turned off - and I think that is an unlikely case.

robertpietzcker:
> The main point is, our prices reporting is much more complicated then it needs to be because of the need to handle these special cases simply because we use two parameters for the same thing in the model: `pm_taxemiMkt` and `pm_taxCO2eq `.

I see the point. 
At the same time I think it would be helpful to not to make everyone else have to deal with an emission price covering several markets if they never use it.

maybe it is not such a big thing, and it can be changed without major effort.

Renato-Rodrigues:
There will be no effect to the model. 
The model is already defined in the higher level of disaggregation (emiMkt). `vm_demFe` and emission equations already include emiMkt as a dimension.

We are talking about a parameter here.
The way it is done now, for every special behavior that somebody creates to `pm_taxCO2eq`, eother in the model or in the reporting, I need to create an special condition to `pm_taxemiMkt` to deal with it, as the latter is only defined at the regipol module. 
If we simply use the same parameter for everything, there is no need at all for such complications. And the parameter to be used must be always the one with higher dimension so we can always derive the others from it.

PR #402, opened 2023-04-26
fix Moving Avg Prices such that periods < cm_startyear are identical to ref run

orichters:
- Let a run have `cm_startyear = 2025`. Then, the `Price|*|Moving Avg` values that are used in [the AR6 mapping](https://github.com/pik-piam/piamInterfaces/blob/master/inst/templates/mapping_template_AR6.csv#L209) for 2020 are `p_average(2020) = (p(2015) + 2p(2020) + p(2025))/4` as calculated in [`remind2::reportPrices()`](https://github.com/pik-piam/remind2/blob/master/R/reportPrices.R#L672) by using [`magclass::lowpass()`](https://github.com/pik-piam/magclass/blob/master/R/lowpass.R). And as the 2025 prices differ between the scenarios, so does the reported 2020 price, which it should not.
- See discussion [here](https://mattermost.pik-potsdam.de/rd3/pl/wwnm331yntduzfg8q5za55x3ca).
- Now, `reportPrices` gets an optional dependency on `gdx_ref` as well, calculates the moving averages for the reference run and uses them for t <- cm_startyear.
- Maybe @fbenke-pik and @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q can have a look whether that makes sense
- Also, I wonder whether there are other places where we have to call `reportPrices` with the `gdx_ref`: I will figure out with @dklein-pik whether convGDX2MIF_REMIND2MAgPIE() as called [in the REMIND reporting](https://github.com/remindmodel/remind/blob/8d3099374865ac88daa02905c7a942f0b805a8d1/scripts/output/single/reporting.R#L47) should also get that added. Not sure whether MAgPIE uses these averages prices…

### DIFF
--- a/R/reportPrices.R
+++ b/R/reportPrices.R
@@ -11,6 +11,8 @@
 #' be created.
 #' @param t temporal resolution of the reporting, default:
 #' t=c(seq(2005,2060,5),seq(2070,2110,10),2130,2150)
+#' @param gdx_ref a GDX object as created by readGDX, or the path to a gdx of the reference run.
+#' It is used to guarantee consistency for Moving Avg prices before cm_startyear
 #'
 #' @return MAgPIE object - contains the price variables
 #' @author Alois Dirnaichner, Felix Schreyer, David Klein, Renato Rodrigues, Falk Benke
@@ -30,7 +32,7 @@
 
 #' @export
 reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
-                         t=c(seq(2005,2060,5),seq(2070,2110,10),2130,2150)) {
+                         t=c(seq(2005,2060,5),seq(2070,2110,10),2130,2150),gdx_ref=NULL) {
 
   ## bind to output object
   if(is.null(output)){
@@ -100,6 +102,7 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
   esm2macro.m    <- readGDX(gdx,name='q35_esm2macro',types="equations",field="m",format="first_found", react = "silent")[, t,]
 
   cm_emiscen     <- readGDX(gdx,name='cm_emiscen',format="first_found")
+  cm_startyear   <- readGDX(gdx,name='cm_startyear',format='simplest')
 
   q37_limit_secondary_steel_share.m <- readGDX(
     gdx, name = 'q37_limit_secondary_steel_share', types = 'equation',
@@ -672,13 +675,20 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
   ## apply lowpass filter to receive moving average prices ----
   out.lowpass <- lowpass(out)
   ## add "Moving Avg" to variable name
-  getNames(out.lowpass) <- paste0(substr(getNames(out.lowpass),
-                                         1,nchar(getNames(out.lowpass))-13),
-                                  "|Moving Avg",
-                                  substr(getNames(out.lowpass),
-                                         nchar(getNames(out.lowpass))-12,
-                                         nchar(getNames(out.lowpass))))
+  getNames(out.lowpass) <- paste0(magclass::unitsplit(getNames(out))$variable,
+                                  "|Moving Avg (",
+                                  magclass::unitsplit(getNames(out))$unit, ")")
 
+  ## reset values for years smaller than cm_startyear to avoid inconsistencies in cm_startyear - 5
+  if (! is.null(gdx_ref)) {
+    priceRef <- try(reportPrices(gdx_ref, output=NULL, regionSubsetList=regionSubsetList, t=t))
+    fixedyears <- getYears(out)[getYears(out, as.integer = TRUE) < as.integer(cm_startyear)]
+    if (! inherits(priceRef, "try-error") && length(fixedyears) > 0) {
+      out.lowpass[, fixedyears, ] <- priceRef[getRegions(out.lowpass), fixedyears, getNames(out.lowpass)]
+    }
+  }
+
+  # bind prices and moving averages together
   out <- mbind(out,out.lowpass)
 
   ## add years before cm_startyear (temporary, can be adapted once prices only calculated after cm_startyear in REMIND code)
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '215655426'
+ValidationKey: '215741367'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'remind2: The REMIND R package (2nd generation)'
-version: 1.107.8
-date-released: '2023-04-20'
+version: 1.107.9
+date-released: '2023-04-26'
 abstract: Contains the REMIND-specific routines for data and model output manipulation.
 authors:
 - family-names: Rodrigues
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: remind2
 Title: The REMIND R package (2nd generation)
-Version: 1.107.8
-Date: 2023-04-20
+Version: 1.107.9
+Date: 2023-04-26
 Authors@R: c(
     person("Renato", "Rodrigues", , "renato.rodrigues@pik-potsdam.de", role = c("aut", "cre")),
     person("Lavinia", "Baumstark", role = "aut"),
--- a/R/convGDX2MIF.R
+++ b/R/convGDX2MIF.R
@@ -68,7 +68,7 @@ convGDX2MIF <- function(gdx, gdx_ref = NULL, file = NULL, scenario = "default",
   message("running reportTechnology...")
   output <- mbind(output,reportTechnology(gdx,output,regionSubsetList,t)[,t,])    # needs output from reportSE
   message("running reportPrices...")
-  output <- mbind(output,reportPrices(gdx,output,regionSubsetList,t)[,t,]) # needs output from reportSE, reportFE, reportEmi, reportExtraction, reportMacroEconomy
+  output <- mbind(output,reportPrices(gdx,output,regionSubsetList,t,gdx_ref = gdx_ref)[,t,]) # needs output from reportSE, reportFE, reportEmi, reportExtraction, reportMacroEconomy
   message("running reportCosts...")
   output <- mbind(output,reportCosts(gdx,output,regionSubsetList,t)[,t,])  # needs output from reportEnergyInvestment, reportPrices, reportEnergyInvestments
   message("running reportTax...")
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # The REMIND R package (2nd generation)
 
-R package **remind2**, version **1.107.8**
+R package **remind2**, version **1.107.9**
 
 [![CRAN status](https://www.r-pkg.org/badges/version/remind2)](https://cran.r-project.org/package=remind2)  [![R build status](https://github.com/pik-piam/remind2/workflows/check/badge.svg)](https://github.com/pik-piam/remind2/actions) [![codecov](https://codecov.io/gh/pik-piam/remind2/branch/master/graph/badge.svg)](https://app.codecov.io/gh/pik-piam/remind2) [![r-universe](https://pik-piam.r-universe.dev/badges/remind2)](https://pik-piam.r-universe.dev/builds)
 
@@ -49,7 +49,7 @@ In case of questions / problems please contact Renato Rodrigues <renato.rodrigue
 
 To cite package **remind2** in publications use:
 
-Rodrigues R, Baumstark L, Benke F, Dietrich J, Dirnaichner A, Führlich P, Giannousakis A, Hasse R, Hilaire J, Klein D, Koch J, Kowalczyk K, Levesque A, Malik A, Merfort A, Merfort L, Morena-Leiva S, Pehl M, Pietzcker R, Rauner S, Richters O, Rottoli M, Schötz C, Schreyer F, Siala K, Sörgel B, Spahr M, Strefler J, Verpoort P (2023). _remind2: The REMIND R package (2nd generation)_. R package version 1.107.8, <URL: https://github.com/pik-piam/remind2>.
+Rodrigues R, Baumstark L, Benke F, Dietrich J, Dirnaichner A, Führlich P, Giannousakis A, Hasse R, Hilaire J, Klein D, Koch J, Kowalczyk K, Levesque A, Malik A, Merfort A, Merfort L, Morena-Leiva S, Pehl M, Pietzcker R, Rauner S, Richters O, Rottoli M, Schötz C, Schreyer F, Siala K, Sörgel B, Spahr M, Strefler J, Verpoort P (2023). _remind2: The REMIND R package (2nd generation)_. R package version 1.107.9, <URL: https://github.com/pik-piam/remind2>.
 
 A BibTeX entry for LaTeX users is
 
@@ -58,7 +58,7 @@ A BibTeX entry for LaTeX users is
   title = {remind2: The REMIND R package (2nd generation)},
   author = {Renato Rodrigues and Lavinia Baumstark and Falk Benke and Jan Philipp Dietrich and Alois Dirnaichner and Pascal Führlich and Anastasis Giannousakis and Robin Hasse and Jérome Hilaire and David Klein and Johannes Koch and Katarzyna Kowalczyk and Antoine Levesque and Aman Malik and Anne Merfort and Leon Merfort and Simón Morena-Leiva and Michaja Pehl and Robert Pietzcker and Sebastian Rauner and Oliver Richters and Marianna Rottoli and Christof Schötz and Felix Schreyer and Kais Siala and Björn Sörgel and Mike Spahr and Jessica Strefler and Philipp Verpoort},
   year = {2023},
-  note = {R package version 1.107.8},
+  note = {R package version 1.107.9},
   url = {https://github.com/pik-piam/remind2},
 }
 ```
--- a/man/reportPrices.Rd
+++ b/man/reportPrices.Rd
@@ -8,7 +8,8 @@ reportPrices(
   gdx,
   output = NULL,
   regionSubsetList = NULL,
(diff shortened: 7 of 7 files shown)


R/reportPrices.R
```diff
@@ -672,13 +675,20 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
...
   ## add "Moving Avg" to variable name
-  getNames(out.lowpass) <- paste0(substr(getNames(out.lowpass),
-                                         1,nchar(getNames(out.lowpass))-13),
-                                  "|Moving Avg",
-                                  substr(getNames(out.lowpass),
-                                         nchar(getNames(out.lowpass))-12,
-                                         nchar(getNames(out.lowpass))))
+  getNames(out.lowpass) <- paste0(magclass::unitsplit(getNames(out))$variable,
+                                  "|Moving Avg (",
+                                  magclass::unitsplit(getNames(out))$unit, ")")
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 680: The double colon operator is superfluous, as `magclass` is imported whole.  If anything, use `@importFrom`.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
> * Let a run have `cm_startyear = 2025`. Then, the `Price|*|Moving Avg` values that are used in [the AR6 mapping](https://github.com/pik-piam/piamInterfaces/blob/master/inst/templates/mapping_template_AR6.csv#L209) for 2020 are `p_average(2020) = (p(2015) + 2p(2020) + p(2025))/4` as calculated in [`remind2::reportPrices()`](https://github.com/pik-piam/remind2/blob/master/R/reportPrices.R#L672) by using [`magclass::lowpass()`](https://github.com/pik-piam/magclass/blob/master/R/lowpass.R). And as the 2025 prices differ between the scenarios, so does the reported 2020 price, which it should not.

It might be that the 2020 prices in files using the AR6 mapping should not differ, but that is not the same as having `Price|*|Moving Avg` variables not being a moving averages in 2020, but something else.
Where else are those moving averages used?

R/reportPrices.R
fbenke-pik: Just a heads-up: magclass::unitsplit is the recommended way to do it (not importFrom). More details in the next RSE newsletter.

orichters:
> It might be that the 2020 prices in files using the AR6 mapping should not differ, but that is not the same as having `Price|*|Moving Avg` variables not being a moving averages in 2020, but something else. Where else are those moving averages used?

The only places the `Moving Avg` variables are used are to the knowledge by myself and reposearch:
- plotting and comparison scripts such as [`AriadneComparison`](https://github.com/remindmodel/remind/blob/develop/scripts/output/comparison/notebook_templates/AriadneComparison.Rmd)
- export templates in piamInterfaces (NAVIGATE, ECEMF, AR6, ARIADNE)

So this is a variable added only for the database export.

R/reportPrices.R
```diff
@@ -672,13 +675,20 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
...
+
+  ## reset values for years smaller than cm_startyear to avoid inconsistencies in cm_startyear - 5
+  if (! is.null(gdx_ref)) {
+    priceRef <- try(reportPrices(gdx_ref, output=NULL, regionSubsetList=regionSubsetList, t=t))
+    fixedyears <- getYears(out)[getYears(out, as.integer = TRUE) < as.integer(cm_startyear)]
+    if (! inherits(priceRef, "try-error") && length(fixedyears) > 0) {
+      out.lowpass[, fixedyears, ] <- priceRef[getRegions(out.lowpass), fixedyears, getNames(out.lowpass)]
+    }
+  }
```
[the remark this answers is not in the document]
orichters: That is what I thought first as well: I want to have the `Moving Avg` prices to be identical to the one calculated by using `gdx_ref`. And say the 2020 value there are calculated by `(p(2015) + 2p(2020) + p(2025))/4`. But the `p(2025)` value for `gdx_ref` and `gdx` differs, so I indeed need this value.

R/reportPrices.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Well, I'm excited for RSE's cogent argument for breaking R namespace conventions.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> So this is a variable added only for the database export.

You cannot infer the reason something has been introduced for in the past from a limited view of what it is used for in the present.

> - plotting and comparison scripts such as [AriadneComparison](https://github.com/remindmodel/remind/blob/develop/scripts/output/comparison/notebook_templates/AriadneComparison.Rmd)
> - export templates in piamInterfaces (NAVIGATE, ECEMF, AR6, ARIADNE)

So this change would affect at least all those other projects, and they should have a say in it.

If memory serves, moving average prices where used instead of real prices because the latter where always weirdly spiky.  That situation might have changed with remindmodel/remind#1238.  Have you checked if you can use proper prices instead?

robertpietzcker:
I would be fine with / in favor of the approach "don't change values in time steps before cm_startyear, even if we are talking about moving averages that theoretically would have to be changed for 100% conceptual consistency".

I have a hard time imagining somebody complaining that "your 2020 prices aren't real moving averages" 
(I can much better imagine "why are your prices so weird and spiky", or "how can prices in the past change?" :-) )

orichters:
FYI:
- the moving averages seem to be introduced in [this PR](https://github.com/pik-piam/remind2/pull/20) by @fschreyer.
- they are generally used in the templates since [this PR](https://github.com/pik-piam/project_interfaces/pull/18) by myself. An alternative would be to revert that.

Renato-Rodrigues:
I am against changing the moving averages calculation to add special rules, as I stated in the mattermost message I sent before.

If you want to post-process any model specific variable (not only prices), you should make very clear what your output reporting variable is doing. 
Price variables are a mess as they are, and adding special rules that violate their own naming conventions only contributes to more confusion in the future.

I started working on a more structured price variable reporting some time ago, following strict rules like:

1. adding the tag `MARGINAL` to prices directly taken from the model without any post-processing. 
2. make sure that additional tags clearly identify the embedded assumption behind specific post-processings, for cases like the `Moving Average` tag.
3. use pure price variable names, without any additional tags, only for our "official" reporting standards that include all necessary post-processing that are necessary to show our results to the outside World.

Meanwhile something like this is not fully implemented, as it is on hold due to other priorities, I would recommend to avoid infringing those principles, and if you have a too specific case, just do any project specific post-processing in project specific scripts.

dklein-pik:
FYI: In the coupling MAgPIE does not use moving average prices but "Price|Carbon (US$2005/t CO2)". The same for the other Kyoto gases.

robertpietzcker:
> FYI: In the coupling MAgPIE does not use moving average prices but "Price|Carbon (US$2005/t CO2)". The same for the other Kyoto gases.

I think this discussion is all about energy prices, which tend to fluctuate in REMIND quite a bit. Carbon prices are (to my knowledge) never averaged

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> I think this discussion is all about energy prices, which tend to fluctuate in REMIND quite a bit. Carbon prices are (to my knowledge) never averaged

This is true, but completely conditional on where in the function [prices are averaged](https://github.com/pik-piam/remind2/blob/f01b223838b5d3018d955d14959f879bf466ec0c/R/reportPrices.R#L672-L682).  Push that down by 100 lines and Carbon prices get their `Moving Avg` buddies.

dklein-pik:
> > FYI: In the coupling MAgPIE does not use moving average prices but "Price|Carbon (US$2005/t CO2)". The same for the other Kyoto gases.
> 
> I think this discussion is all about energy prices, which tend to fluctuate in REMIND quite a bit. Carbon prices are (to my knowledge) never averaged

It was just a reaction to Olli's question at the top.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Have you checked if you can use proper prices instead?

Well, still positively alpine in parts …
image

Renato-Rodrigues:
May I ask why you completely ignored in your "possible options" the structured rules that I mentioned above, and that makes use of the already existent `MARGINAL` tag?

I would say that the best alternative instead would be to follow the already started process of:

1. Using the already existent "MARGINAL" tag to reflect any price directly taken from the model. This tag is only necessary to be added to prices that require some kind of post-processing. 
2. Adapt, if necessary, the moving average variables to use these marginal price variables instead.
3. Use the clear price variable names to store our "release ready" values, after all post-processing is done. This includes adding any necessary moving average assumption; your change on the fixed years enforcement; and additional changes like for example enforcing minimum and maximum acceptable price values for corner cases when the solver marginals are at extreme values (this rule should be applied before the moving average post-processing in cases it is necessary).

This proposed solution follow the structure already in place, do not break existent mappings that use the current moving average calculation, and make sure that in the future we can use price variable names close to the templates used in projects without the need for selecting prices with special tags. We would always have our "clean" price naming convention reserved for our best interpretation of what would be the correct prices to report externally.

orichters:
Sorry, @Renato-Rodrigues, I honestly don't know why my mind forgot about your comment when I was thinking about the alternatives. Weird. I actually think your proposal is the most convincing one.

Did you already prepare something related to this process that I can build on?

As an alternative, I would suggest to rename all the `Price|*` variables for which now a `Price|*|Moving Avg` is calculated to `Price|*|Marginal`, calculate the `Price|*|Moving Avg` as lowpass of the Marginals, and use `Price|*` for the adapted moving average? All other prices would stay unchanged… What about compareScenarios, MAgPIE coupling etc., should we use the Marginals there or the new reporting-"proof" prices?

robertpietzcker:
> What about compareScenarios, MAgPIE coupling etc., should we use the Marginals there or the new reporting-"proof" prices?

I think the compScen should always at least show non-smoothed prices so we always see how spiky prices are currently. (if somebody wants to see the averaged as well, fine with me)

orichters:
Okay, implemented a first attempt of what I described above:

Some statistics, comparing old and new results of `reportPrices`:
- Without fixing, old variables that differ from new Marginal ones: 0
- With fixing, old variables that differ from new Marginal ones: 0
- With fixing, Moving Avg variables that differ from old ones: 0
- Without fixing, Moving Avg that differ from reporting variables: 0
- With fixing, Moving Avg that differ from reporting variables > 2020: 0
- With fixing, Moving Avg that do not differ from reporting variables in 2020: 629
- With fixing, Moving Avg that differ from reporting variables < 2020: 100

Doesn't look too bad. I think, in theory, the values below 2020 should not differ, if the fixing was perfect, but 100 data points is not too bad. If affects:
``` 
[1] "Price|Final Energy|Buildings|Gases|Natural Gas|Total LCOE|Moving Avg"
 [2] "Price|Final Energy|Buildings|Liquids|Oil|Fuel Cost|Moving Avg"
 [3] "Price|Final Energy|Buildings|Liquids|Oil|Total LCOE|Moving Avg"
 [4] "Price|Final Energy|Industry|Gases|Natural Gas|Total LCOE|Moving Avg"
 [5] "Price|Final Energy|Industry|Liquids|Oil|Fuel Cost|Moving Avg"
 [6] "Price|Final Energy|Industry|Liquids|Oil|Total LCOE|Moving Avg"
 [7] "Price|Final Energy|Industry|Solids|Coal|Total LCOE|Moving Avg"
 [8] "Price|Final Energy|Industry|Solids|Moving Avg"
 [9] "Price|Final Energy|Transport|Gases|Natural Gas|Total LCOE|Moving Avg"
[10] "Price|Final Energy|Transport|Liquids|Diesel|Fuel Cost|Moving Avg"
[11] "Price|Final Energy|Transport|Liquids|Diesel|Total LCOE|Moving Avg"
[12] "Price|Final Energy|Transport|Liquids|Petroleum|Fuel Cost|Moving Avg"
[13] "Price|Final Energy|Transport|Liquids|Petroleum|Total LCOE|Moving Avg"
[14] "Price|Secondary Energy|Liquids|Fossil|Moving Avg"
[15] "Price|Secondary Energy|Liquids|Moving Avg"
```
The difference is < 3% except for 
```
    region variable                                          period  MovA    Rep
 49 REF    Pr|FE|Ind|Solids|Moving Avg                         2005  3.94  0.701
 55 REF    Pr|FE|Ind|Solids|Moving Avg                         2010  3.40  2.32
```
If somebody wants to investigate, mif files and scripts are lying around in `/p/tmp/oliverr/remind2-prices`.

orichters:
We have to check inside remind2:
- inst/markdown/compareScenarios2/cs2_02_macro.Rmd -> use `|Marginal`?
- R/reportCharts.R -> Marginals or not?
- R/reportCosts.R -> Marginals or not?

outside remind2:
- piamInterfaces (all) templates? -> use `Price|*` without `|Moving Avg`
- [remindmodel: AriadneComparison](https://github.com/remindmodel/remind/blob/develop/scripts/output/comparison/notebook_templates/AriadneComparison.Rmd) -> deleting `|Moving Avg`.
- [mip: plotstyle.csv](https://github.com/pik-piam/mip/master/inst/extdata/plotstyle.csv): make sure also `|Marginal` values are plotted correctly

orichters:
@dklein-pik wrote me about reportCosts:

> I think I would first try to stay as close as possible to the unchanged results, if the resulting quality of the results calculated by reportCosts allows it, i.e. they do not also vary strongly. The marginal is currently read in reportCosts for some quantities directly from the model and not obtained via reportPrices, as e.g. here https://github.com/pik-piam/remind2/blob/e66ab2bd4d65a0f9c6800a00d3d1e15660799fd2/R/reportCosts.R#L132. This would have to be changed if we do not want to use the marginal, so that we are consistent within reportCosts.

bs538:
After some discussion with Oliver I'd like to suggest a further addition to the price postprocessing: 

Issue: For non-SSP2 scenarios, we currently have wild FE price fluctuations in 2025 (sometimes several 100%), which are caused by the fixing of non-SSP2 scenarios to SSP2-NDC until 2020. We discussed it in more detail in the Macro channel: https://mattermost.pik-potsdam.de/rd3/pl/g156uh5mgt8upjcxmgzocug7fe
For an example of results, see most FE prices here: /p/projects/piam/SHAPE_runs/coupled_2023-05-03_final/remind/compScen-SSP-SDP-PkBudg650-newWithTaxInc-2023-05-23_13.28.57-REMIND-MAgPIE.pdf

As a *temporary* fix, while we come up with a new way of handling the fixing (hopefully as part of the switch to NPi calibrations), we could implement one of the following options:

- A: for all non-SSP2 scenarios, replace 2025 (=cm_startyear) prices with average of 2020 and 2030. This needs to be done prior to calculating the moving average (akin to the current removal of negative prices) to prevent the spikes propagating into 2020 and 2030

- B: replace only prices that fluctuate beyond a current threshold, e.g. set p_tstart = min(max(0.5 * pavg, p_tstart), 1.5 * pavg), where pavg = (p_(tstart-5) + p_(tstart+5)) / 2 (suggestion by Oliver). This would also happen prior to the moving average.

- C: modify the moving average calculation so that 2025 prices are discarded (through adjusting the weights) - either for all non-SSP2 scenarios (as option A) or if they fluctuate beyond a threshold (as option B). This has the advantage that cleaning and moving average happen in one go (no double-averaging for 2025 price), but sounds a bit fiddly as one needs to modify the weighting scheme used in magclass::lowpass() for specific years and scenarios

In my view, Option A would be preferable, as it targets only the prices we know to be affected, and does so without having to specify an arbitrary threshold how much fluctuation is OK.

@Renato-Rodrigues @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q @LaviniaBaumstark any thoughts?

orichters:
@bs538: Another alternative would be to add something like that after [this line](https://github.com/pik-piam/remind2/pull/402/files#diff-9079799de4d750220602928460d4d1998c56d82cbb46b1f6b1c95405768aec36R693).
```
out.reporting[, cm_startyear, ] <- 0.5 * ( priceRef[getRegions(out), cm_startyear, getNames(out)] +
                                           out.reporting[, cm_startyear, ] )
```
So the fixing would not be in time, but rather you avoid diverging too much from the reference run.

bs538:
@orichters : thanks, interesting idea. But I think using the averaging in time domain would be more appropriate in this case, because we do actually expect a strong deviation from the reference run when the carbon price comes in in cm_startyear (especially for ambitious policy scenarios). I quickly looked at FE price changes from 2020 to 2025 in SSP2-PkBudg650 scenario (which has no fixing/adjustment issue), and here we also have price jumps of 50+ % for some FE|Sector|Carriers.

bs538:
> I mean, that is the general problem: I don't think we can easily differentiate legitimate price jumps from the others.
> 
> So indeed maybe the best temporary option is to identify from the gdx whether it is no SSP2 setting, and apply one of the options above. Modifying the weighting scheme seems tedious to me, so I would go with either A or B, depending on whether you actually like to keep part of the price changes or not.

Fully agree on the general problem. That's why any sort of threshold approach is problematic. So I would pick Option A (because we know that non-SSP2 scenarios have a price problem in 2025).



> The question is whether this is best done here or in some sort of post-processing. I have no strong opinion on that.

As we already fix price issues (such as negative prices, and smoothing out spikes) here, this seems to be the right place to add this further fix. If we do it in a separate post-processing, we need to go back to the Rawdata / Marginal prices, fix the problem there, and then re-apply the MovingAvg, so a lot of duplication.



> And I think we should put this issue high on the agenda to fix these problems at the root (if that is the root): [remindmodel/remind#1068](https://github.com/remindmodel/remind/issues/1068)

Again fully agree.

bs538:
@orichters : thanks!

@0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q @Renato-Rodrigues @fbenke-pik quick review would be appreciated as I would like to use this improvement to the price reporting (both general and in particular [d3b1a54](https://github.com/pik-piam/remind2/pull/402/commits/d3b1a5439430b3f558d997d2057d35d3c4de1e6b) ) for final project scenarios quickly. thanks!

fbenke-pik:
Sorry, I don't think if i can be of help when it comes to validating rules for reporting variables. Whenever I extend the reporting (adding/adjusting reporting variables), it happens under supervision of felix, robert or renato. I just do the tech, but don't know anything about the logic behind the calculations.

Renato-Rodrigues (approved):

orichters:
Find attached a number of plots where I compare
- the old reporting
- the new reporting without any reference (nothing should have changed)
- the new reporting with a base run to which it is fixed until 2020
- the reference run

It all looks good from my side. (I moved the lines a bit left and right to make everything visible)

[alldata.pdf](https://github.com/pik-piam/remind2/files/11576302/alldata.pdf)

PR #406, opened 2023-05-19
removing dependency of regionSubsetList

Renato-Rodrigues:
BugFix to remove dependency of regionSubsetList in the reportEmi function.


orichters:
Thanks, looks good!

robertpietzcker (approved):

Issue #408, opened 2023-05-26
variables with unit NA


orichters:
Warning (test-convGDX2mif.R:120:5): Test if REMIND reporting is produced as it should and check data integrity These variables and units contain NA:
- Services and Products|Transport|non-LDV|Sales|apCarDiT (NA)
- Services and Products|Transport|non-LDV|Sales|apcarDiEffH2T (NA)
- Services and Products|Transport|non-LDV|Sales|apcarDiEffT (NA)
- Services and Products|Transport|non-LDV|Sales|uedit (NA)
- Services and Products|Transport|non-LDV|Stock|apCarDiT (NA)
- Services and Products|Transport|non-LDV|Stock|apcarDiEffH2T (NA)
- Services and Products|Transport|non-LDV|Stock|apcarDiEffT (NA)
- Services and Products|Transport|non-LDV|Stock|uedit (NA)

robertpietzcker:
sorry for not reacting to the messages earlier. 
Is it a problem if they are NA? non-LDV is a variable that is very hard to put into a useful unit....

orichters:
Those specific variables are now ignored.

PR #410, opened 2023-05-31
use explicit gdx_refprices in convGDX2MIF, ignore LDV variables with NA unit

orichters:
- needs adaptation in remind such that the gdx_refprices is passed to convGDX2MIF.
- I was a bit unsure whether to rename `gdx_ref` to `gdx_refpolicycosts`  in order to be consistent, but I'm a bit afraid of breaking stuff… @LaviniaBaumstark / @Renato-Rodrigues, would you prefer that?
- accept `NA` in some LDV units (see #408), as suggested by @robertpietzcker
- improve explanations in reportPrices
-

### DIFF
--- a/R/convGDX2MIF.R
+++ b/R/convGDX2MIF.R
@@ -12,6 +12,7 @@
 #' @param scenario scenario name that is used in the *.mif reporting
 #' @param t temporal resolution of the reporting, default:
 #' t=c(seq(2005,2060,5),seq(2070,2110,10),2130,2150)
+#' @param gdx_refprices reference-gdx for < cm_startyear, used for fixing the prices to this scenario
 #' @author Lavinia Baumstark
 #' @examples
 #'
@@ -22,9 +23,10 @@
 #' @importFrom magclass mbind write.report
 
 convGDX2MIF <- function(gdx, gdx_ref = NULL, file = NULL, scenario = "default",
-                        t = c(seq(2005, 2060, 5), seq(2070, 2110, 10),
-                              2130, 2150)) {
-   # Define region subsets
+                        t = c(seq(2005, 2060, 5), seq(2070, 2110, 10), 2130, 2150),
+                        gdx_refprices = NULL) {
+
+  # Define region subsets
   regionSubsetList <- toolRegionSubsets(gdx)
   # ADD EU-27 region aggregation if possible
   if("EUR" %in% names(regionSubsetList)){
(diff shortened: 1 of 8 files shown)


orichters:
Ok, so if `gdx_refpolicycost` is not specified, it defaults to `gdx_ref`, which allows to be backward-compatible. In REMIND, the situation should never happen that `gdx_refpolicycost` does not exist while `gdx_ref` is there because if `path_gdx_refpolicycost` is empty, it falls back automatically to `path_gdx_ref`.

orichters:
I now set `gdx_refpolicycost = gdx_ref` in the function description, so you can overwrite it as you like, but if it is not provided, it defaults to `gdx_ref`:
```
> z <- function(x = 3, y = x) return(paste("x:", x, "/ y:", y))
> z(3, 5)
[1] "x: 3 / y: 5"
> z(3)
[1] "x: 3 / y: 3"
```

R/convGDX2MIF.R
```diff
@@ -5,26 +5,28 @@
 #'
 #'
 #' @param gdx a GDX as created by readGDX, or the file name of a gdx
-#' @param gdx_ref reference-gdx for policy costs, a GDX as created by readGDX, or the file name of a gdx
+#' @param gdx_ref reference-gdx for < cm_startyear, used for fixing the prices to this scenario
```
LaviniaBaumstark, line 8: there is no `gdx_ref`argument anymore

R/convGDX2MIF.R
```diff
@@ -5,26 +5,28 @@
...
 #' @examples
 #'
-#' \dontrun{convGDX2MIF(gdx,gdx_ref,file="REMIND_generic_default.csv",scenario="default")}
+#' \dontrun{convGDX2MIF(gdx,gdx_refpolicycost,file="REMIND_generic_default.csv",scenario="default")}
 #'
 #' @export
 #' @importFrom gdx readGDX
 #' @importFrom magclass mbind write.report
 
 convGDX2MIF <- function(gdx, gdx_ref = NULL, file = NULL, scenario = "default",
```
LaviniaBaumstark, line 25: why do we need gdx_ref still for price correction?

bs538:
Sounds good to me. I understand the user workflow for calculating policy costs is still the same after this, right?
- selecting refpolicycost in scenario config OR manually through ex-post output.R -> policyCosts
- (if rerunning reporting): automatically using right reference

orichters:
> Sounds good to me. I understand the user workflow for calculating policy costs is still the same after this, right?
> 
>     * selecting refpolicycost in scenario config OR manually through ex-post output.R -> policyCosts
> 
>     * (if rerunning reporting): automatically using right reference

Yes, that remains unchanged. I thought about renaming the `gdx_ref` function parameter in [reportPolicyCosts](https://github.com/pik-piam/remind2/blob/master/R/reportPolicyCosts.R#L8) as well to `gdx_ref_policycost`, but then I thought it is unnecessary, may cause scripts to fail and the title of the function sufficiently clearly suggests that we are talking about policy costs here, I guess.

R/convGDX2MIF.R
orichters: Because prices are Moving Averages, and if `cm_startyear` = 2025, and then the policy scenario calculates its 2020 price values, this takes into account the 2025 value which differs from its reference scenario, and so the reported prices would differ in 2020 even though they should not. See the long debates in https://github.com/pik-piam/remind2/pull/402

R/convGDX2MIF.R
orichters: that is not correct, we still need `gdx_ref`…

R/convGDX2MIF.R
LaviniaBaumstark: ah, thanks

LaviniaBaumstark (approved):

orichters:
For the record: Had a chat with @Renato-Rodrigues, he is fine with the changes in reportPrices.

PR #411, opened 2023-06-03
rescaling non-energy reporting values to match EU-27 data

Renato-Rodrigues:
This is a short term solution for the wrongly reported EU-27 non-energy use.

The previous numbers are being re-scaled to match Eurostat 2020 non-energy use data. 
The same scaling factor is applied to all years and energy carriers for EU-27 countries.


robertpietzcker (approved):
As usual, I don't fully get the R code, but if the output is adjusted to the expected value - great!

PR #417, opened 2023-06-22
make sure cs2 shows FE prices with reporting versions before Rawdata PR

orichters:
- the old script dropped too many variables if called with old reporting versions before https://github.com/pik-piam/remind2/pull/402, and particularly messed up everything if the scenarios were done with different reportings. Therefore, do the selection per scenario.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '217006800'
+ValidationKey: '217193130'
 AcceptedWarnings:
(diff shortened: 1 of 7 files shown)


Renato-Rodrigues (approved):
If you tested with both a recent and an older version of results, I have nothing to add to the PR.

orichters:
Test is here: `/p/tmp/oliverr/compScen-2C_comparison_apr_to_june_2023-2023-06-19_17.04.28-H12.pdf`

PR #419, opened 2023-06-23
Update reporting after refactoring the CDR module (portfolio)

katarkow:
Report CDR after introducing a new CDR module (one realization `portfolio`, in which options are chosen by switches instead of module realizations) [REMIND #1154](https://github.com/remindmodel/remind/pull/1154).

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '217728081'
+ValidationKey: '217876120'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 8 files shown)


strefler (commented):
I think fehes is missing in reportCosts, otherwise looking good!

strefler (approved):

katarkow:
@strefler Indeed, fehes is missing in reportCosts.R, thanks for noticing! However, it also has been the case for the previous version of the code + fehes doesn't really show up as a source of energy for DAC so it doesn't affect our results. I'll open a separate issue to address this and merge the PR without fehes.

PR #422, opened 2023-07-03
add industry carbon captured by subsector

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


fschreyer (approved):
Looks good. Two comments

1.) I personally prefer things like simply repeating the code of assigning the ``pm_IndstCO2Captured`` to the output variable because its more readable than the automated routine. Probably a matter of taste and coding habits. It would be good if you commented somewhere what you do with this section overall, which is just mapping the entries of ``pm_IndstCO2Captured`` to the appropriate variable names. 

2.) Other Industry does not have carbon capture? Or should it be added?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> 1.) I personally prefer things like simply repeating the code of assigning the `pm_IndstCO2Captured` to the output variable because its more readable than the automated routine.

I wholeheartedly disagree.  Repeating 48 almost identical statements turns code into chequered wallpaper that makes your eyes glaze over.  And this table arrangement keeps the salient differences between the individual variables – which subsets are selected – in plain view.
Added some comments, though.

> 2.) Other Industry does not have carbon capture? Or should it be added?

Other Industry does not have CCS in REMIND.

PR #427, opened 2023-07-04
fix aggregation of intensive variable `Carbon Management|Share of Stored CO2 from Captured CO2 (%)`

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


fschreyer (approved):
Looks good.

Issue #438, opened 2023-08-10
#402 breaks price reporting for certain scenario combinations


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
This
https://github.com/pik-piam/remind2/blob/cafb0ddd1999c63a12fb9e4c771c4ee9f521a6e1/R/reportPrices.R#L705-L707
erroneously presumes that all variables in `out` are always present in `out.reporting`, which is not the case for
```
Price|Final Energy|Buildings|Gases|Natural Gas|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Buildings|Liquids|Oil|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Industry|Gases|Natural Gas|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Industry|Liquids|Oil|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Industry|Solids|Coal|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Transport|Gases|Natural Gas|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Transport|Liquids|Diesel|Carbon Price Component (US$2005/GJ)
Price|Final Energy|Transport|Liquids|Petroleum|Carbon Price Component (US$2005/GJ)
```
when `gdx_ref` refers to a run where `p_priceCO2`, and thus any "Carbon Price Component" is zero, resulting in 
```
Error in .dimextract(x, k, 3, pmatch = pmatch, invert = invert) : 
  subscript out of bounds …
```

PR #439, opened 2023-08-11
wrap reporting function calls in confGDX2MIF() inside tryCatch() and continue on failed reportings

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
This way, if some reporting function is buggy (current example: #438), the rest of the reporting can proceed.  If users are not interested in the failed reporting, they can carry on. Otherwise they are no worse off then before.
Additional reporting functions are simply inserted into a list along with arguments they might need.


R/convGDX2MIF.R
```diff
@@ -1,128 +1,172 @@
...
-  output <- NULL
-  message("running reportMacroEconomy...")
-  output <- mbind(output,reportMacroEconomy(gdx,regionSubsetList,t)[,t,])
-  message("running reportTrade...")
-  output <- mbind(output,reportTrade(gdx,regionSubsetList,t)[,t,])
-  message("running reportPE...")
-  output <- mbind(output,reportPE(gdx,regionSubsetList,t)[,t,])
-  message("running reportSE...")
-  output <- mbind(output,reportSE(gdx,regionSubsetList,t)[,t,])
-  message("running reportFE...")
```
fbenke-pik, line 48: I think these messages are helpful to see the progress (at least when running reporting locally) and should not be removed.

fbenke-pik (commented):
I think introducing try-catch blocks around the reporting functions makes sense.

I'd like to get some further opinions on the refactoring, though. Michaja introduced elegant helpers to reduce redundant code portions. But potentially this could make it harder for people to understand what is going on and extend the reporting. 

Since this is the central reporting function, I'd like to be sure that people are cool with this approach before merging. 

@fschreyer @LaviniaBaumstark @orichters @Renato-Rodrigues @pweigmann 
Any thoughts?

fbenke-pik:
I personally am fine with this approach, if there are no objections by others.

orichters:
I'm ok with the change. Some things are a bit less obvious now, in particular which of the functions prefer to get `output` passed. I think in the old style, it was obvious that the order of the functions mattered as they got passed the `output` object. Now, this is blurred completely in the helper functions and I think this is problematic. If you move `reportExtraction` above `reportCosts`, then `output` that is passed to `reportCosts` is not `NULL`, so `reportCosts` will not call `reportExtraction`, and it will fail quite intransparently later, I guess.

But maybe a simple way to get transparency about this dependencies is by explicitly adding a comment `reportCosts, # needs reportExtraction, reportPrices, reportEnergyInvestments` or something like that behind the respective line.

orichters:
And I was confused again by the `gdx_ref` and `gdx_ref = gdx_refpolicycosts`. Not sure whether it was the right decision in [PR #410](https://github.com/pik-piam/remind2/pull/410) to prioritize backwards-compatibility over clarity.

R/convGDX2MIF.R
```diff
@@ -1,128 +1,172 @@
...
+  .get_reporting_function <- function(reporting_function_name) {
+    # Get the reporting function by name from the remind2 package, print an
+    # expressive error message should that fail.
+    tryCatch(
+      getFunction(reporting[[1]], where = getNamespace('remind2')),
+      error = function(e) {
+        known_error_message <- paste0('^no function .', reporting[[1]],
+                                      '. found$')
+        if (grepl(known_error_message, e[['message']])) {
+          stop('Function foo() not defined in package remind2.')
```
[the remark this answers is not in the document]
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Ooops.

R/convGDX2MIF.R
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: > see the progress

As in "is this running or did this crash"?
I was [under the impression](https://github.com/pik-piam/remind2/commit/218738e369fe73123939731ce1a65fc04e9c53d8) this was added to figure out where it crashed if/when it crashed, which is now handled explicitly.
Anyhow, added messages (that can be turned off).

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> Some things are a bit less obvious now, in particular which of the functions prefer to get `output` passed. I think in the old style, it was obvious that the order of the functions mattered as they got passed the `output` object. Now, this is blurred completely in the helper functions and I think this is problematic.  If you move `reportExtraction` above `reportCosts`, then `output` that is passed to `reportCosts` is not `NULL`, so `reportCosts` will not call `reportExtraction`, and it will fail quite intransparently later, I guess.

That would have been true before as well.
```
output <- NULL
output <- reportMacroEconomy(gdx)
output <- mbind(output, reportCosts(gdx, output))
output <- mbind(output, reportExtraction(gdx))
```
will also have `is.null(output)` and fail on some `subscript out of bounds`.  Compared to before, there is now at least [a comment telling people that the order of calls does matter](https://github.com/pik-piam/remind2/pull/439/files#diff-c4e2cb978caee670f1b02a44264a8c7b59215e7b174ed32ad75e5024eae8cba1R51-R52).

P.S.: The proper way to handle this would be to check for the presence of required variables in the `output` passed into the reporting functions.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> And I was confused again by the `gdx_ref` and `gdx_ref = gdx_refpolicycosts`. Not sure whether it was the right decision in [PR #410](https://github.com/pik-piam/remind2/pull/410) to prioritize backwards-compatibility over clarity.

I added a comment on that.

orichters:
A (maybe stupid) idea:
```
reportPrices <- function(gdx, output = c("reportPE", "reportSE", …), …)
```

Then later check not `is.null(output)` but `is.character(output)` or something similar.

And your script can use `formals(reportPrices)$output` to check whether the necessary reportings are above in the list.

Renato-Rodrigues:
I must confess that I am not a big fan of using lists and for loops to define parameters in sequential function calls.
I would prefer if we create instead a report wrapper function that includes the verbose and try catch additional codes, but keep using the sequential-one-line-for-each-report-call structure instead.

I am fine anyway if you want to follow this path, but I think it is much easier to follow the single line function calls structure on debugging the code than entering in loops for that.

fschreyer:
I agree with Renato's view.

PR #440, opened 2023-08-11
remove Final Energy|Share of renewables in gross demand|Rough estimation as it is wrong

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@bs538, this also removes `FE|Share of renewables in gross demand` from "SDG 7 clean" in the compare scenarios PDF.  No idea if that variable ever existed, but it does not now in `remind2` and was silently ignored by the compare scenarios scripts.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '218157654'
+ValidationKey: '218199520'
 AcceptedWarnings:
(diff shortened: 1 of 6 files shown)


robertpietzcker (approved):

bs538:
Fine for me to remove that variable, I am not using it.

bs538 (approved):

Issue #425, opened 2023-07-03
`Secondary Energy|Electricity|Share of renewables in gross demand|Estimation`


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> It might arise from a mistaken understanding of "gross demand" that I had some years ago, namely that it was more connected to the final energy demand. In fact, "gross electricity demand" is just the sum of electricity generation plus imports (excluding output from pump hydro storage that has no natural inflows).
> 
> accordingly, the best approximation for "Secondary Energy|Electricity|Share of renewables in gross demand|Estimation" would be
> ```R
>   100 
> * ( output[,,"SE|Electricity|Non-Biomass Renewables (EJ/yr)"] 
>   + output[,,"SE|Electricity|Biomass (EJ/yr)"]
>   )
> / ( output[,,"SE|Electricity (EJ/yr)"]
>   + output[,,"SE|Electricity|Net Imports (EJ/yr)"]
>   )
> ```

[[↑]](https://github.com/pik-piam/remind2/issues/420#issuecomment-1618390468)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
@robertpietzcker 
That will not work, as `SE|Electricity|Net Imports` is not available for all realisations of the `trade` module.
https://github.com/pik-piam/remind2/blob/cafb0ddd1999c63a12fb9e4c771c4ee9f521a6e1/R/reportSE.R#L350-L365

robertpietzcker:
Ok. 

I know very little R magic, so a simple question, @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q  
Would it be possible to simply always report the variable `SE|Electricity|Net Imports`, and if there is no entry, set it to 0 (because if there is no variable in REMIND, then clearly imports were 0)?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Sure can do.

PR #441, opened 2023-08-11
adjust Secondary Energy|Electricity|Share of renewables in gross demand|Estimation calculation

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
close https://github.com/pik-piam/remind2/issues/425

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '218199520'
+ValidationKey: '218252535'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


robertpietzcker (approved):
Thanks!

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I collapsed `SE|Electricity|Net Imports (EJ/yr)` to zero instead of adding more zeros to every .mif file.  Seems more economical.

PR #444, opened 2023-08-24
refactor variable summation check in tests

fbenke-pik:
This PR is meant to replace https://github.com/pik-piam/remind2/pull/274, using piamInterfaces::checkSummations to test summations of a mif file generated with convGDX2MIF.


fbenke-pik:
One open question: Anastasis' PR moved the check into convGDX2MIF, so that it is executed everytime the reporting is run, and not only as part of a unit test. Let me know if you prefer this approach over just doing it manually / in the test for an example mif.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> One open question: Anastasis' PR moved the check into convGDX2MIF, so that it is executed everytime the reporting is run, and not only as part of a unit test. Let me know if you prefer this approach over just doing it manually / in the test for an example mif.

How long does the test take, relative to the conversion?

orichters:
I would like to have it executed automatically
- in the remind2 tests
- while running [`reporting.R`](https://github.com/remindmodel/remind/blob/develop/scripts/output/single/reporting.R#L139) in remindmodel.
No opinion whether it should be part of the function itself, though.

fbenke-pik:
> How long does the test take, relative to the conversion?
I think it should not slow down reporting noticeably, the check itself doesn't take longer than 40 sec and in relation to running the other reporting functions, it does not take long. 

It more depends on people's view on warnings due to mismatches in summations.
Warnings are printed for variables that exceed a certain threshold. If we want to get people's attention, that's fine. If nobody does anything about it and it is just clutter in the log, then it might not be a good idea.

fbenke-pik:
Let me do further research on these discrepancies listed above to assure they are justified.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
Having once spend a couple of days fixing all errors from misformed `put` commands in REMIND, when the output was still written using GAMS, I'm in favour of including the test in `convGDX2MIF()`.

The whole point of the `|+|` notation is to test whether the numbers do add up or not.
I think the 1 % limit on relative differences is huge.  These numbers should match exactly.  `all.equal()` uses `tolerance = sqrt(.Machine$double.eps)`, which is 1.5e-8.  If we have an explanation for why these numbers don't match up, we should adapt the reporting.
And -40 % is pretty much "WTF?!" territory!

Come to think of it, we should run this test – and report the results – in the automated model tests.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

fbenke-pik:
> Come to think of it, we should run this test – and report the results – in the automated model tests.

Do you mean instead of convGDX2MIF, or in addition to convGDX2MIF?

> all.equal() uses tolerance = sqrt(.Machine$double.eps), which is 1.5e-8

Happy to adapt this. The tolerance I chose is somewhat arbitrary anyways because i don't know what is acceptable or not in that context.

orichters:
@fbenke-pik: I think if you set `template = NULL`, you will have a cleaner output without the
```
              <
                + 
                + 
                + 
                + 
                + 
```
on the right side.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> > Come to think of it, we should run this test – and report the results – in the automated model tests.
> 
> Do you mean instead of convGDX2MIF, or in addition to convGDX2MIF?

I think `convGDX2MIF()` should always run the `'extractVariableGroups'` test by default and warn if things don't add up.  I don't see a scenario where we should accept any deviation in the numbers we generate from our own model's output.  And the AMT should include any warnings from the test in their report, so that we get flagged if something does not add up anymore.

> > all.equal() uses tolerance = sqrt(.Machine$double.eps), which is 1.5e-8
> 
> Happy to adapt this. The tolerance I chose is somewhat arbitrary anyways because i don't know what is acceptable or not in that context.

In this context – checking our own postprocessing – absolute values should match to within machine precision (and relative differences are inconsequential).

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
> @fbenke-pik: I think if you set `template = NULL`, you will have a cleaner output without the
> 
> ```
>               <
>                 + 
>                 + 
>                 + 
>                 + 
>                 + 
> ```
> 
> on the right side.

I would not have guessed that from the parameter discription:
> `template` mapping template to be loaded

:thinking:

orichters:
>> template mapping template to be loaded

> 🤔

Yes, because the whole code was written with a different purpose in mind: You were meant to feed in data in `AR6` or `NAVIGATE` format – and then the script uses the mapping template to find out which `remind2` variables are related. Now, Falk uses `template = "AR6"` (because this is the default), but the data is not in AR6 format and therefore it shows nonsense. But yes, we might need to adapt the function description now that its features were extended.

fbenke-pik:
Right, the documentation of the `template` parameter should be improved.

@orichters we might also want to parametrize the tolerance for the human-readable output. Right now, it is hard coded to 
`abs(!!sym("reldiff")) >= 1, abs(!!sym("diff")) >= 0.001`, which apparently is too soft in this context.

fbenke-pik:
Right, `.checkSummationsSummary` which generates the output was written with another purpose in mind. We have two options: 
1) deactivate it for this use case after all and do analysis on the output of checkSummations
2) enhance it for this use case

@orichters what do you prefer and in case of option 2) do you want to do it, or should I?

tests/testthat/test-convGDX2mif.R
```diff
@@ -10,61 +10,6 @@
...
-        collapse = "\n"
-      ))
-    }
-  }
-}
-
-# please add variable tests below
-checkIntegrity <- function(out, gdxPath = NULL) {
-  dt <- rmndt::magpie2dt(out)
-  barspace <- grep("[\\| ]{2}", unique(dt[["variable"]]), value = TRUE)
```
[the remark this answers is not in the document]
fbenke-pik: Done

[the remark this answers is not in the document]
fbenke-pik: This should only yield a new result when variables have been renamed, so I think as part of lucode::buildLibrary should be sufficient, no? So dont see a reason to call it from `convGDX2MIF`. But I can make it available as a helper function and export it if you see yourself using it in other places as well.

fbenke-pik:
I moved the summations check to convGDX2MIF so the warnings are printed every time the reporting is run. For now, a warning is thrown whenever the absolute value for a parent variable exceeds 1.5e-8

R/convGDX2MIF.R
```diff
@@ -115,7 +115,17 @@ convGDX2MIF <- function(gdx, gdx_ref = NULL, file = NULL, scenario = "default",
...
 
+  sumChecks <- piamInterfaces::checkSummations(
+    mifFile = output, outputDirectory = NULL,
+    summationsFile = "extractVariableGroups",
+    absDiff = 1.5e-8, relDiff = 1e-8, roundDiff = FALSE
+  ) %>% filter(abs(!!sym("diff")) >= 1.5e-8)
+
+  if (nrow(sumChecks) > 0) {
+    warning("Summation checks have revealed some gaps! Run piamInterfaces::checkSummations")
+  }
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 128: I wonder if it would be worthwhile to write the `checkSummations()` output to a file directly.
People will see that somewhere in the log file, but for most REMIND developers it won't be obvious what they should use `piamInterfaces::checkSummations()` on.
Just dumping this into some file would not cost us anything (kilobytes on gigabytes per run), but would greatly facilitate both debugging and ascertaining the validity of model output for projects.
```
  if (nrow(sumChecks) > 0) {
    if (is.null(file)) {
      warning("Summation checks have revealed some gaps! Run piamInterfaces::checkSummations")
    }
    else {
      summation_errors_file <- sub('(\\.[^.]+)$', '_summation_errors\\1', file)
      warning("Summation checks have revealed some gaps! See file ",
              summation_errors_file)
      write.table(sumChecks, summation_errors_file, quote = FALSE, sep = ';')
    }
  }
```

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Come to think of it, it would also be easier to just return the summation errors as an attribute (much like [`readr::problems()`](https://readr.tidyverse.org/reference/problems.html)), instead of telling people to run some extra code.
:point_right:  https://github.com/fbenke-pik/remind2/pull/1

R/convGDX2MIF.R
```diff
@@ -115,7 +115,17 @@ convGDX2MIF <- function(gdx, gdx_ref = NULL, file = NULL, scenario = "default",
   output <- add_dimension(output,dim=3.1,add = "model",nm = "REMIND")
   output <- add_dimension(output,dim=3.1,add = "scenario",nm = scenario)
 
+  message("Checking integrity of created MIF...")
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 118: Not sure about this message.  This is checking only one aspect of data integrity.  And since we expect the reporting to be error-free, it would only add a meaningless line to the log file

R/convGDX2MIF.R
orichters: I agree. Except you could use `dataDumpFile` functionality of `checkSummations()` for that.

R/convGDX2MIF.R
fbenke-pik: Happy to delete it :)

R/convGDX2MIF.R
fbenke-pik: Right, so unless @orichters wants to refactor checkSummations to only write to a file if some intolerable gaps were discovered, we go with Michaja's approach for now.

Issue #418, opened 2023-06-22
Cleanup FE Prices in cs2


orichters:
After adding FE price variables in https://github.com/pik-piam/remind2/pull/402 and fixing the bug that omitted all Energy prices in https://github.com/pik-piam/remind2/pull/417, these [`walk` calls for FE Prices](https://github.com/pik-piam/remind2/blob/master/inst/markdown/compareScenarios2/cs2_02_macro.Rmd#L168-L194) in `cs2_02_macro.Rmd` of compareScenarios2 create way too many plots. It was 18 pages in January, now it is > 200. I would appreciate if you could replace this code to select the variables you actually look at. Maybe @fschreyer, @Renato-Rodrigues?

fschreyer:
I think we don't need the line plots and bar plots for the distinction of FE Prices in  ``Total LCOE``, ``Fuel Cost``, ``Transport and Distribution``, ``Carbon Price Component``, ``Other Taxes``. They can be all removed. If somebody wants to know about it, they can work based on the mif files.

PR #451, opened 2023-09-05
cleanup FE prices in cs2

orichters:
- As noted in #418, PR #402 added hundreds of pages to cs2 with `Price|Final Energy|` categories.
- I remove all these automatic "walks" and replace them with some hand-picked variables, as @fschreyer suggested I also removed the LCOE variables
- Adding some users, if you think others may need it please add them as reviewers.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '219007455'
+ValidationKey: '219049404'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '218987850'
+ValidationKey: '219007455'
 AcceptedWarnings:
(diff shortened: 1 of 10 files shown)


Renato-Rodrigues:
Could you share two pdfs with the situation before and after the change so we can take a look if we spot any important chart missing after the change?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
Fine by me.
Would you mind merging https://github.com/orichters/remind2/pull/1 first?  I _really_ prefer to have quantities in the proper order.

orichters:
@Renato-Rodrigues: You can look at `/p/projects/remind/modeltests/remind/output/default-AMT_2023-09-01_22.06.04/comp_with_default-AMT_2023-08-04_22.06.07.pdf`. Pages 98 – 366 will be gone, and replaced by

- `Price|Final Energy|Buildings|*`
- `Price|Final Energy|Transport|*`
- `Price|Final Energy|Industry|*`
 
with `*` from `c("Electricity", "Gases", "Heat", "Hydrogen", "Liquids", "Solids")` (if they exist)

Renato-Rodrigues (approved):

PR #452, opened 2023-09-06
fix calculation of gross emissions in GHG emissions variables and complete emissions barplots in cs2 summary part

fschreyer:
This fixes the calculation of gross emissions for GHG variables in analogy to #394, which is necessary for consistent summations with gross variables and CDR variables (gross emissions + CDR = net emissions). Moreover, I added some more explanation to this calculation. I also added some variables to the sectoral gross + CDR barplots in the summary section of cs2. I furthermore checked the summations of those barplots using ``piamInterfaces::checkSummations`` and they are now fine for a recent gdx. 

Not sure about ongoing processes here, but we as discussed with Simon we might need a redefinition of gross variables if we want to avoid them being negative when synfuels are captured and stored. Maybe discuss this in a dedicated group with people working on emissions/synfuels/CDR at some point.


amerfort (approved):
Thanks Felix for taking care of it! I can only judge the remind2 part and that one looks good. In general I think the explanation of why some gross variables can still turn negative could be improved, as even though I know the issue I had to wrap my head around the sentences for a couple of minutes :)

fschreyer:
I tried to improve the explanation. Thanks!

PR #461, opened 2023-09-12
Set FE|Non-energy use variables to zero if they have negative values

pweigmann:
Due to they way they are calculated, some variables like ``FE|w/o Non-energy Use|Industry|Chemicals|Liquids|+|Hydrogen (EJ/yr)`` could become negative. This is avoided by setting them to zero in case they are `< 0`.


fbenke-pik:
Could this adjustment cause problems with the variables not summing up as indicated by the plus notation?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):
> Due to they way they are calculated, some variables […] could become negative.

Then we should fix how these variables are calculated.

Can you give a specific example (including gdx file) where a variable is computed as having a negative value?

pweigmann:
> > Due to they way they are calculated, some variables […] could become negative.
> 
> Then we should fix how these variables are calculated.
> 
> Can you give a specific example (including gdx file) where a variable is computed as having a negative value?

This is one example: `/p/tmp/schreyer/Modeling/remind/ariadne/output/Bal_2023-08-21_15.26.32/fulldata.gdx`. From what I understood from the code, we are calculating non-energy variables based on fixed values from 2020 - so depending on our run it would be quite difficult to ensure that we never go negative when subtracting. Happy about input for a better way to calculate it.

Also, as it is a temporary calculation, this fix was thought as also just a temporary fix to avoid negative values in our project reportings.



> Could this adjustment cause problems with the variables not summing up as indicated by the plus notation? If so, are we ok with that? (the reporting would yield a warning in this case as we now have summation checks based on plus notation) Or should instead the calculation of this variable be done in a different way?

That is true, I didn't think about it. Another argument for a different calculation.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

PR #476, opened 2023-10-02
Adding wind offshore and onshore primary energy variables

Renato-Rodrigues:
Adding wind offshore and onshore primary energy variables to be used in project reportings.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '219760608'
+ValidationKey: '219780240'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


robertpietzcker (approved):
looks good. 
will this crash when people turn off wind offshore? ($IFTHEN.WindOff %cm_wind_offshore% == "0")?

Renato-Rodrigues:
> looks good. will this crash when people turn off wind offshore? ($IFTHEN.WindOff %cm_wind_offshore% == "0")?

It shouldn't, but just to be on the safe side I added an extra check to be double secure.
https://github.com/pik-piam/remind2/pull/476/commits/3e9d7624669d558f61ba78d4cf5b330736743bbe

PR #479, opened 2023-10-06
Add biogasc reporting variables

merfort:
This PR adds two biogas reporting variables that were missing so far.


robinhasse (approved):
Looks good! Thanks @merfort

PR #481, opened 2023-10-12
Bugfix of previously added SE variable for specific GDX files

fschreyer:
Bugfix for #480 to work with this GDX: 

https://rse.pik-potsdam.de/data/example/remind2_test-Ariadne_fulldata.gdx

I tested the PR only with https://rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_fulldata.gdx and I presume``buildLibrary`` does so, too. I would suggest this GDX file to also feature ``make test``. What do you think, @LaviniaBaumstark?

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '219976439'
+ValidationKey: '220029684'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


orichters (approved):
this gdx is featured in `make test`, that is why it failed for me. The question is rather why the test did not fail in your case…

PR #482, opened 2023-10-18
calculate Price|Carbon from pm_taxCO2eqSum

orichters:
- instead of `pm_pvpRegi`
- requires some changes in remindmodel
- fixes some inconsistencies between run and reference run https://github.com/remindmodel/development_issues/issues/220

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '220254080'
+ValidationKey: '220461780'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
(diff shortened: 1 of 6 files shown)


LaviniaBaumstark (commented):
looks good, I need a revision number for REMIND GAMS code

orichters:
Do you see any problem with the [R/readPVP.R](https://github.com/pik-piam/remind2/pull/482/files#diff-95c401c6b6b3c2966097aa8de2d918cd045698350bd7a6dc1036f0f94cf3c14b) stuff? I have no idea where this is actually used…

Sure, I will run `make build`…

LaviniaBaumstark:
I thing, most of the read* functions in remind2 are from the time before we moved the whole reporting to R. They might we outdated :-)

orichters:
1.122.0 it is…

PR #488, opened 2023-11-22
Update LCOE reporting script

fschreyer:
This PR provides a clean up and some extensions to the LCOE reporting script. 

Relevant additions are:

* add adjustment cost to marginal LCOE calculation
* add marginal LCOE calculation for CO2 storage technology (ccsinje)
* provide two versions of marginal LCOE: 1.) with fuel price and carbon price from time step for which LCOE is calculated (``time step prices``). 2.) with an intertemporally weighted average of fuel prices and carbon prices over the lifetime of the plant (``intertemporal price``). Previously, only 1.) had been calculated. The second formulation aims to better capture the intertemporal dimension of investment decisions in the model.  

This may be interesting also for @aodenweller and @JakobBD.


fbenke-pik (approved):

JakobBD:
Looks great! 
Adjustment costs in LCOE are tricky; a few points on that: 

-  Increasing deltaCap does not only increase adjustment costs in the current time step, but also reduces them in the following one. This could be accounted for by computing the derivative of adjustment costs wrt `deltaCap(ttot-1)` in `core/postsolve` and subtracting this (discounted) next timestep's derivative $\frac{\partial\text{costAdj(ttot+1)}}{\partial\text{deltaCap(ttot)}}$. 
- Adjustment costs can be very large, dominating LCOE
- The marginal values imho don't always make sense in an economic interpretation, especially in combination with bounds. For example, `ccsinje` bounds lead to negative marginal adjustment costs when approaching those bounds. It's still valuable to understand how remind makes its decisions, but maybe not so good for externally presented figures. How do we want to treat this? Are adjustment costs actual economic values, or just virtual ones we use to tweak the model?

fschreyer:
Thanks for your points. Good to keep that in mind.

aodenweller:
Hi Felix, I checked your forward-looking LCOE calculations by writing down the equations from the standard LCOE definition again and everything looks good. Thanks!
The only question I have is why you fix r to 5.1 % instead of using the endogenous interest rate `p_r` from REMIND? But the difference should be negligible.

fschreyer:
> Hi Felix, I checked your forward-looking LCOE calculations by writing down the equations from the standard LCOE definition again and everything looks good. Thanks! The only question I have is why you fix r to 5.1 % instead of using the endogenous interest rate `p_r` from REMIND? But the difference should be negligible.

Ok, interesting. I actually did not check with the LCOE definition. However, I am still not entirely sure how to interpret the intertemporal LCOE components as the resulting total LCOE sometimes tend to match even less the REMIND prices as the time-step LCOE in my analyses. But I did not check in detail the correlation with investment decisions e.g. in ``vm_deltaCap``. That's why I added both now and we might investigate further which fits which purpose.

cchrisgong:
@fschreyer I added p_r some time ago as an endogenous interest rate.. I'm also thinking whether we should use this in LCOE script to stay consistent

JakobBD:
Yes, we absolutely should.

PR #489, opened 2023-11-23
Add reporting and calculations for feedstocks and plastics

mellamoSimon:
This PR adds new (endogenously calculated) feedstock and plastic variables when available, including necessary calculations for emissions and final energy demand. Everything should be hidden behind if clauses to secure backward compatibility. 
It also includes some new plots for these variables in cs2.


R/reportFE.R
```diff
@@ -516,9 +539,9 @@ reportFE <- function(gdx, regionSubsetList = NULL,
     setNames(out[,,"FE|Transport|++|ESR (EJ/yr)"], "FE|Transport|w/o Bunkers (EJ/yr)"),
     setNames(out[,,"FE|Transport|++|Outside ETS and ESR (EJ/yr)"], "FE|Transport|Bunkers (EJ/yr)")
   )
-  out <- mbind(out,
-               setNames(out[,,"FE (EJ/yr)"] - out[,,"FE|Transport|Bunkers (EJ/yr)"], "FE|w/o Bunkers (EJ/yr)")
-  )
+ # out <- mbind(out,
+ #              setNames(out[,,"FE (EJ/yr)"] - out[,,"FE|Transport|Bunkers (EJ/yr)"], "FE|w/o Bunkers (EJ/yr)")
+#  )
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 544: You still need this?

R/reportFE.R
```diff
@@ -1344,10 +1368,16 @@ reportFE <- function(gdx, regionSubsetList = NULL,
...
+  #### Non-energy Use Reporting ----
+
+ ### temporary (!) industry non-energy use reporting
   # note: only for REMIND-EU SSP2
 
-  if ("DEU" %in% getRegions(vm_prodFe) & indu_mod == 'subsectors') {
+  if ("DEU" %in% getRegions(vm_prodFe) & indu_mod == 'subsectors' & is.null(vm_demFENonEnergySector)) {
+
+      # some initializations required for building library with dplyr operations below
+      encar <- data <- value <- value_subsectors <- SSP <- Value_NonEn <- encar <- region <- period <- NULL
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 1379: This is bad practice.  Use tidyselect syntax (i.e. `'encar'`) or the `.data` pronoun instead.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (changes requested):

R/reportFE.R
mellamoSimon: arg, I think Felix had deleted that before, and now I had wrongly kept it when pulling. I just deleted it. thank you! maybe @fschreyer would like to have a look?

R/reportFE.R
mellamoSimon: commenting it out was a mistake. thanks.

R/reportEmi.R
```diff
@@ -93,6 +93,11 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
   }
   se_sol <- c("sesofos", "sesobio")
 
+  # sets of hydrocarbon SE carriers
+  entySeFos <- readGDX(gdx, "entySeFos")
+  entySeBio <- readGDX(gdx, "entySeBio")
+  entySeSyn <- readGDX(gdx, "entySeSyn")
+
   # SE carriers by origin
```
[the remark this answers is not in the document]
mellamoSimon: now it's only using `entySEfos`, `entySEbio`, and `entySEsyn`. was that what you meant?

R/reportEmi.R
```diff
@@ -906,6 +1099,24 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
...
+  if (!is.null(vm_demFENonEnergySector)) {
+    # calculate chemical process emissions from feedstocks treatment
+    EmiFac_NonEn.co2 <- collapseDim(pm_emifacNonEnergy[,,"co2"])
+    EmiProcess_Feedstocks <- pm_emifacNonEnergy[,,"co2"] * vm_demFENonEnergySector[,,getNames(EmiFac_NonEn.co2)]
+    out <- mbind(out,
+                 setNames(dimSums(vm_feedstockEmiUnknownFate, dim=3)* GtC_2_MtCO2,
+                          "Emi|CO2|Feedstocks unknown fate (Mt CO2/yr)"),
+                 setNames(dimSums(vm_incinerationEmi, dim=3)* GtC_2_MtCO2,
+                          "Emi|CO2|Plastics incineration (Mt CO2/yr)")
+                 )
```
fschreyer, line 1115: These two variables do not really conform to our emissions category taxonomy. Let's maybe think again how to label them and fit into the structure. Are they process or energy-related emissions?

``Emi|CO2|Plastics incineration`` could be incineration for energy purposes (IPCC 1A1) e.g. for power generation or incineration without energy recovery (IPCC 5C). Is that a correct understanding? If so, the first category we would account under ``Emi|CO2|Energy|Supply|+|Electricity/Heat``. The second we would account under ``Emi|GHG|+++|Waste``. (Or we would add a ``Emi|CO2|+|Waste``). 

Concerning the Feedstock emissions with "unknown fate", we would also need a decision of whether these emissions are energy-related (IPCC 1) or belong to some other sector. 

Maybe it is also good to check again whether the IAM community has already defined something on this (I fear not, though) e.g. by looking at the variable definitions of the NAVIGATE template.

R/reportEmi.R
```diff
@@ -1155,6 +1375,47 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
...
+  )
+
+  # carbon management: fate of plastics
+  out <- mbind(out,
+
+            # carbon in plastics that are incinerated
+            setNames(dimSums(vm_incinerationEmi, dim = 3) * GtC_2_MtCO2,
+                    "Carbon Management|Materials|Plastics|++|Incineration (Mt CO2/yr)"),
+            # carbon in plastics with other fate
+            setNames(dimSums(vm_nonIncineratedPlastics, dim = 3) * GtC_2_MtCO2,
```
fschreyer, line 1402: What about ``vm_feedstockEmiUnknownFate`` here? Are they part of ``vm_nonIncineratedPlastics``? Or are they not accounted as CO2 going into materials?

R/reportEmi.R
```diff
@@ -539,8 +623,12 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
...
                           "Emi|CO2|Energy|Demand|+|Buildings (Mt CO2/yr)"),
                # industry emissions: fe carrier emissions - industry CCS
-               # (carbon captured by industry used for CCU is still accounted as industry emissions as only account negative emissions for the CO2 provider if carbon gets stored)
-               setNames((dimSums(EmiFeCarrier[, , "indst"], dim = 3) - dimSums(vm_emiIndCCS[, , emiInd37_fuel]* p_share_CCS, dim = 3) ) * GtC_2_MtCO2,
+               # (carbon captured by industry used for CCU is still accounted as industry emissions
+               # as only account negative emissions for the CO2 provider if carbon gets stored)
+               setNames((dimSums(EmiFeCarrier[, , "indst"], dim = 3)
+                         - dimSums(vm_emiIndCCS[, , emiInd37_fuel]* p_share_CCS, dim = 3)
+                         # substract synthetic and biogenic carbon contained in industrial feedstocks
+                         - dimSums(plastic_CDR, dim=3)) * GtC_2_MtCO2,
```
fschreyer, line 631: Just a quick refresher for my memory. Here, we only substract the plastic carbon from biogenic and synthetic sources and not the carbon from fossil sources because fossil feedstock carbon is not accounted as emissions in ``EmiFeCarrier`` in the first place. Correct?

fschreyer: Also this means that we account all (non-fossil) carbon stored in materials as negative emissions for industry.

R/reportFE.R
```diff
@@ -1344,10 +1368,13 @@ reportFE <- function(gdx, regionSubsetList = NULL,
...
 
-  ### temporary (!) industry non-energy use reporting
+  #### Non-energy Use Reporting ----
+
+ ### temporary (!) industry non-energy use reporting
   # note: only for REMIND-EU SSP2
 
-  if ("DEU" %in% getRegions(vm_prodFe) & indu_mod == 'subsectors') {
+  if ("DEU" %in% getRegions(vm_prodFe) & indu_mod == 'subsectors' & is.null(vm_demFENonEnergySector)) {
+
```
fschreyer, line 1377: Yeah, I think our non-energy use reporting for ARIADNE can actually be removed once this is merged. Although you are right it is maybe safe to still keep it here and I will check out our new Germany scenarios with the feedstocks implementation afterwards and then delete this part, if everything looks good.

fschreyer (commented):
Hey Simon, looks good overall! :)

Just some clarification questions and comments from my side.

R/reportFE.R
```diff
@@ -1653,25 +1680,204 @@ reportFE <- function(gdx, regionSubsetList = NULL,
...
+                               "FE|Transport|Liquids|+|Biomass (EJ/yr)")
+
+  fe.vars.woBunkers.syn <- c(  "FE|Liquids|+|Hydrogen (EJ/yr)",
+                               "FE|Transport|Liquids|+|Hydrogen (EJ/yr)")
+
+
+  # add FE w/o non-energy use variables if available
+  if ("FE|Non-energy Use (EJ/yr)" %in% getNames(out)) {
+
+    # bunker correction for distinction of fossil, biomass, hydrogen-based liquids
```
fschreyer, line 1872: Shouldn't we also add something here to also get ``FE|w/o Bunkers|w/o Non-energy Use``. So the top-level FE variable without both?

R/reportFE.R
fschreyer: Well, I guess it will given you an error when building, if you do not have it, right? But leave it for now if you did not introduce it. Probably the whole section can be deleted soon anyways (see below).

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

R/reportEmi.R
mellamoSimon: For the records, we fixed it with the new category `Emi|CO2|Energy|+|Waste`, under which we created: `Emi|CO2|Energy|Waste|+|Plastics Incineration` and `Emi|CO2|Energy|Waste|+|Feedstocks unknown fate`

R/reportEmi.R
```diff
@@ -2204,9 +2528,12 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
 
 
 
-  ## 8. Emissions w/o non-energy use ----
-  # (Note: The non-energy use variables are so far only available for REMIND-EU runs and industry fixed_shares)
+  # 8. Ad-hoc fix for emissions w/o non-energy use and Aggregation to global and regional values  ----
+
+if (is.null(vm_demFENonEnergySector) && (module2realisation["industry", 2] == "fixed_shares")) {
```
fbenke-pik, line 2533: This condition breaks reportEmi for some project-specific reportings (for example Ariadne) and will be reverted.

PR #490, opened 2023-12-05
add `Price|Carbon|Imported` for ELEVATE

orichters:
- add `Price|Carbon|Imported` for ELEVATE
- make sure it is backwards-compatible
- make sure data missing but listed in `int2ext` for global aggregation is not leading to fails anymore

### DIFF
--- a/R/reportPrices.R
+++ b/R/reportPrices.R
@@ -786,6 +786,25 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
                                 "Price|Carbon|AggregatedByGrossCO2 (US$2005/t CO2)")) # AggregatedByEmiGHGGross
   }
 
+  tradeFossil <- c("pecoal", "pegas", "peoil")
+  p21_tau_Import <- readGDX(gdx, name = "p21_tau_Import", react = "silent")[, t, tradeFossil]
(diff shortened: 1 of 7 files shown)


R/reportPrices.R
```diff
@@ -1015,7 +1032,9 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
   map <- data.frame(region=getRegions(out),world="GLO",stringsAsFactors=FALSE)
   tmp_GLO <- new.magpie("GLO",getYears(out),getNames(out),fill=0)
 
-  for (i2e in names(int2ext)){
+  int2ext <- int2ext[intersect(names(int2ext), getNames(out))] # select only data that exists
```
sophiefuchs, line 1035: not familiar with this. would be great if @pweigmann could check this

sophiefuchs (approved):

RahelMA (approved):

R/reportPrices.R
pweigmann: looks good!

pweigmann (approved):

orichters:
Fixed another error. If none of the `if` clauses evaluates to TRUE, it runs `setNames(0)` and that fails because this is not a magclass object. Now, initalize with 0 * pm_taxCO2eq and this works both on newer Elevate gdx files as old NGFS files.

PR #492, opened 2023-12-08
adding average LCO-ccsinje and adjustment cost bugfixes

tabeado:
- added average levelized cost of ccsinje and cost from electricity demand for ccsinje in report.LCOE.
- bugfix of average adjustment cost in report.LCOE
- bugfix of "capital cost w/ adjCost" for ccsinje, dac, and some transport techs in report.Technology


tabeado:
@robertpietzcker @JakobBD

R/reportLCOE.R
```diff
@@ -126,8 +127,9 @@ reportLCOE <- function(gdx, output.type = "both"){
  p47_taxCO2eq_AggFE <- readGDX(gdx,"p47_taxCO2eq_AggFE", restore_zeros=F, react = "silent")
 
  ## variables
- v_directteinv <- readGDX(gdx,name=c("v_costInvTeDir","vm_costInvTeDir","v_directteinv"),field="l",format="first_found")[,ttot,]
- v_directteinv_wadj <- readGDX(gdx,name=c("o_avgAdjCostInv"),field="l",format="first_found")[,ttot,]
+ v_directteinv <- readGDX(gdx,name=c("v_costInvTeDir","vm_costInvTeDir","v_directteinv"),field="l",format="first_found")[,ttot,] ## Total direct Investment Cost in Timestep
+ #v_directteinv_wadj <- readGDX(gdx,name=c("o_avgAdjCostInv"),field="l",format="first_found")[,ttot,] ## average adjustment cost per unit installed in period
+ v_adjteinv <- readGDX(gdx,name=c("v_costInvTeAdj"),field="l",format="first_found")[,ttot,] ## total adjustment cost in period
```
cchrisgong, line 132: maybe the R variable can also be called "v_costInvTeAdj"? normally I think the convention is to go with the REMIND variable name unless it is not possible.. (like v_directteinv above this)

R/reportLCOE.R
```diff
@@ -237,12 +239,28 @@ reportLCOE <- function(gdx, output.type = "both"){
...
+      fuels <- c("peoil","pegas","pecoal","peur", "pebiolc" , "pebios","pebioil",
+                 "seel","seliqbio", "seliqfos","seliqsyn", "sesobio","sesofos","seh2","segabio" ,
+                 "segafos","segasyn","sehe")
+      pm_PEPrice <- readGDX(gdx, "pm_PEPrice", restore_zeros = F) 
+      pm_SEPrice <- readGDX(gdx, "pm_SEPrice", restore_zeros = F)
+      Fuel.Price <- mbind(pm_PEPrice,pm_SEPrice )[,,fuels]*1e12 # convert from trUSD2005/TWa to USD2005/TWa
+
+    # calculate secondary Fuel cost for ccsinje 
+      te_annual_secFuel_cost <- new.magpie(getRegions(te_inv_annuity),ttot_from2005, "ccsinje" , fill=0)  
+      te_annual_secFuel_cost[,,"ccsinje"] <- setNames(-pm_prodCouple[,,"ccsinje"] * Fuel.Price[,,"seel"] * vm_co2CCS[,,"ccsinje.1"], "ccsinje")
```
cchrisgong, line 261: I am probably missing something, but why negative here? (-pm_prodCouple)

R/reportLCOE.R
```diff
@@ -237,12 +239,28 @@ reportLCOE <- function(gdx, output.type = "both"){
...
+                 "seel","seliqbio", "seliqfos","seliqsyn", "sesobio","sesofos","seh2","segabio" ,
+                 "segafos","segasyn","sehe")
+      pm_PEPrice <- readGDX(gdx, "pm_PEPrice", restore_zeros = F) 
+      pm_SEPrice <- readGDX(gdx, "pm_SEPrice", restore_zeros = F)
+      Fuel.Price <- mbind(pm_PEPrice,pm_SEPrice )[,,fuels]*1e12 # convert from trUSD2005/TWa to USD2005/TWa
+
+    # calculate secondary Fuel cost for ccsinje 
+      te_annual_secFuel_cost <- new.magpie(getRegions(te_inv_annuity),ttot_from2005, "ccsinje" , fill=0)  
+      te_annual_secFuel_cost[,,"ccsinje"] <- setNames(-pm_prodCouple[,,"ccsinje"] * Fuel.Price[,,"seel"] * vm_co2CCS[,,"ccsinje.1"], "ccsinje") 
+          # units: pm_prodCouple[,,"ccsinje"] = [TWa/GtC]; Fuel.Price = [USD2005/GtC]; vm_co2CCS= [GtC]; te_annual_secFuel_cost = [USD2005]
```
cchrisgong, line 262: fuel price has unit USD2005/TWa, right?

R/reportLCOE.R
```diff
@@ -319,8 +337,8 @@ reportLCOE <- function(gdx, output.type = "both"){
 
  # calculate total ccsinjection cost for all techs
  total_ccsInj_cost <- dimReduce(te_annual_inv_cost[getRegions(te_annual_OMF_cost),getYears(te_annual_OMF_cost),"ccsinje"] +
-                                                     te_annual_OMF_cost[,,"ccsinje"])
- # distribute ccs injection cost over techs
+                                                     te_annual_OMF_cost[,,"ccsinje"] + te_annual_secFuel_cost[,,"ccsinje"])
+ # distribute ccs injection cost over techs => this does not include adjustment cost incurred! => what is the reasoning behind it?
```
cchrisgong, line 341: don't quite understand this comment, do you mean why is adjustment cost not added as a OMF cost? What is the unit of total_ccsInj_cost?

Btw I don't quite get the formula here; shouldn't the total cost = annuitized (investment cost + other fixed costs (OMF, adjCost)) + running cost (energy cost, OMV cost, CO2 tax)... It looks like OMF and energy cost are here, but the others(?)

R/reportLCOE.R
```diff
@@ -445,7 +466,18 @@ LCOE.avg <- NULL
               setNames(te_annual_co2_cost[,,pe2se$all_te]/total_te_energy[,,pe2se$all_te],
                        paste0("LCOE|average|",pe2se$all_enty1,"|",pe2se$all_te, "|supply-side","|CO2 Cost")),
               setNames(te_curt_cost[,,pe2se$all_te],
-                       paste0("LCOE|average|",pe2se$all_enty1,"|",pe2se$all_te, "|supply-side","|Curtailment Cost"))
+                       paste0("LCOE|average|",pe2se$all_enty1,"|",pe2se$all_te, "|supply-side","|Curtailment Cost")),
+              #### Carbon Transport and storage ("ccsinje")
+              setNames(te_annual_inv_cost[,ttot_from2005,"ccsinje"]/
+                         vm_co2CCS_tCO2[,,"ccsinje.1"],
+                       paste0("LCOx|average|","cco2_stored|","ccsinje","|supply-side", "|Investment Cost")),
```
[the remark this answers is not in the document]
cchrisgong: why do we give "cco2_stored" in the variable name? I thought usually for readability we have something more like "Captured CO2|Injected CO2 |Supply side|..." etc

cchrisgong (changes requested):
Maybe would also be nice to have a trial run (or output LCOE csv available - or its screenshot), so a CCS expert can comment on the costs (not an expert here :P) maybe Jess or Anne?

R/reportLCOE.R
tabeado: I agree in principle, but wanted to stick with the convention used in this part of the script. "v_costInvTeDir" turns into "v_directteinv", and I just applied the same logic. What did you mean saying "(like v_directteinv above this)"?

R/reportLCOE.R
tabeado: It says in line 249 that negative values mean own consumption, positive values mean coupled product. As I want positive cost, I need to multiply. 
I can adjust the line with the units if you want to have a more detailed documentation.

R/reportLCOE.R
tabeado: Good catch, will adjust

R/reportLCOE.R
tabeado: I put this in the commit accidentally. The point is that only ccs injection cost without adjustment cost are distributed across technologies. I just wondered why there is no version that includes adjustment cost as these can be very substantial

tabeado: The general formula for total cost should be (and is):
annuitized (investment cost + adjCost)  + OMF cost (don't need to be annuitized) + running cost (energy cost, OMV cost, CO2 tax).
The ccs cost only has investment cost + OMF cost + now also energy cost
Adjustment cost are missing, that was precisely the point of my comment

R/reportLCOE.R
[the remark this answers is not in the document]
tabeado: I called it "cco2_stored" because the convention for the other items in this dimension is pe2se$all_te. In my understanding, "cco2" would be the remind element corresponding to that. However, it is not about cco2 (all captured co2) but only the part of it that is stored, thus the additional "_stored"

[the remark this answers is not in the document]
tabeado: about LCOx vs LCOCS: I think that is a general question of how we want to call "LCO other than E" in this script. There are plans to also add LCO of CDR options. Do we want to give all of them their own abbreviation? Or do we want to have a generic LCOx and then filter according to the technologies, as it is the case with LCOE?

tabeado:
You can compare results here: /p/tmp/tabeado/remind2_clean/reportLCOE_adjustment 
And you may recall that we found these bugs and added ccsinje precisely because I am reviewing the cost for ccsinje right now :)

cchrisgong (approved):
Thanks for the nice edits! I approve this change :) Happy holiday~ :3

PR #494, opened 2023-12-11
Add nash analysis markdown

fbenke-pik:
This PR adds a new output script `nashAnalysis` rendering various plots to analyze nash convergence.
@Renato-Rodrigues This PR also removes the no longer maintained output script `dashboard`,  as there are no plans to fix and use it again.


robertpietzcker (approved):
I can't comment on the R code, but if it allows creating the convergence html from the output scripts interface in a REMIND run, great!

PR #499, opened 2023-12-15
adding detailed FE plot (deep el paper figure 1)

cchrisgong:
adding the deepEl FE detailed graph

image


fbenke-pik (approved):

fbenke-pik (changes requested):
Please increment library version.

PR #501, opened 2024-01-04
fix errors and warnings in reportEmi

fbenke-pik:
This PR increments the library version, as it was not done in https://github.com/pik-piam/remind2/pull/489
In addition, warnings and errors caused by this PR in reportEmi are fixed as well.

### DIFF
--- a/R/reportEmi.R
+++ b/R/reportEmi.R
@@ -11,7 +11,7 @@
 #' @param t temporal resolution of the reporting, default:
 #' t=c(seq(2005,2060,5),seq(2070,2110,10),2130,2150)
 #'
-#' @author Felix Schreyer...
+#' @author Felix Schreyer, Simón Moreno Leiva
 #' @examples
 #' \dontrun{
 #' reportEmi(gdx)
@@ -26,7 +26,6 @@
 
 
 reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(2005, 2060, 5), seq(2070, 2110, 10), 2130, 2150)) {
-
 
   # emissions calculation requires information from other reporting functions
   if (is.null(output)) {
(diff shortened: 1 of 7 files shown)


tabeado (approved):

mellamoSimon:
ups, too slow I guess... still, I think is good that @fschreyer takes a look. thank you for fixing my mess though, Falk! I thought the tests would fail if I didn't do that and I probably forgot to check manually. thanks again!

R/reportEmi.R
```diff
@@ -2536,10 +2535,9 @@ if (!is.null(vm_plasticsCarbon)) {
...
 
-if (is.null(vm_demFENonEnergySector) && (module2realisation["industry", 2] == "fixed_shares")) {
- # (Note: The non-energy use variables are so far only available for REMIND-EU runs and industry fixed_shares)
+
+  # (Note: The non-energy use variables are so far only available for REMIND-EU runs and industry fixed_shares)
   # TODO: add non-energy use variables for all regionmappings and sector realizations
-  #Note (SM): I'm not sure if I got these notes so I created the condition above to try to make sure that this will work anyways
 
   # Note: Non-energy use emissions should not be confused with process emissions. Non-energy use emissions are emissions/carbon flow of FE carriers which are used as feedstocks in industry.
   if ("FE|Non-energy Use|Industry (EJ/yr)" %in% getNames(output) &&
```
fschreyer, line 2543: What would be important is that the version with ``vm_demFENonEnergySector`` does not enter this bracket. We do not need emissions variables corrected for non-energy use in Simon's implementation anymore. However, it looks like it does not enter here because Simon named the variables slightly differently e.g. ``FE|Non-energy Use|+|Industry (EJ/yr)`` instead of ``FE|Non-energy Use|Industry (EJ/yr)``. However, for clarity of the code it would advisable to add a ``is.null(vm_demFENonEnergySector)`` condition.

PR #503, opened 2024-01-10
Update reporting after renaming v33_emi to vm_emiCdrTeDetail

tabeado:
Update reporting after renaming v33_emi to vm_emiCdrTeDetail in https://github.com/remindmodel/remind/pull/1517

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '222568458'
+ValidationKey: '222644604'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


amerfort:
The reporting is supposed to have a certain backwards-compatibility, right? So we would probably have to have a check if v33_emiCDR still exists (because the respective gdx is a little older) and do the calculation the old way in this case. @katarkow would you agree?

amerfort:
Oh I totally missed that in the PR, I'm so sorry! Thanks for taking care of this and thanks for explaining!

amerfort (approved):
Thank you very much for taking care of this! Looks good to me.

PR #513, opened 2024-01-29
adjust calculation in nash markdown

fbenke-pik:
This affects the calculation of the undiscounted good price.


robertpietzcker (approved):
Thanks!

PR #519, opened 2024-02-05
adding overlapped regional tech cost plots as a function of cumulative capacity for learning tech

cchrisgong:
image

note: regions with small cumulative capacity benefits from global spillover under the default scenario of one price globally and global learning


cchrisgong:
could be interesting to @PhilippVerpoort

fbenke-pik (approved):

PR #521, opened 2024-02-06
if path_gdx_refpolicycost empty, report Policy costs as 0

orichters:
- if `path_gdx_refpolicycost` is empty, which usually means you are in a current policies scenario, report Policy Costs as zero to avoid to simply not have these variables in your submission.
- Set `dataDumpFile` to NULL for checkSummations, as a joint file is written later


fbenke-pik (approved):

johanneskoch94 (approved):
I think this should be fine! Thanks :)

PR #523, opened 2024-02-06
Restore condition for project-specific emissions reporting

mellamoSimon:
This restores a condition designed to enable the reporting of industry emissions from feedstocks with an ad-hoc ex-post fix designed for the Ariadne project. Used when feedstock flows are not available (older remind versions).
This PR is NOT intended to fix the inconsistencies in emissions accounting. That will come in this one: https://github.com/pik-piam/remind2/pull/530 .

### DIFF
--- a/R/reportEmi.R
+++ b/R/reportEmi.R
@@ -479,8 +479,6 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
   }
 
 
-
-
   # calculate total energy supply and demand co2 emissions
   out <- mbind(out,
 
@@ -1121,7 +1119,9 @@ if (!is.null(vm_plasticsCarbon)) {
                   - (1 - p_share_CCS) * (-vm_emiCdrTeDetail[, , "dac"])
                   # deduce co2 captured by industrial processes which is not stored but used for CCU (synfuels)
                   # -> gets accounted in industrial process emissions
-                  - vm_emiIndCCS[, , "co2cement_process"]*(1-p_share_CCS)) * GtC_2_MtCO2,
+                  - vm_emiIndCCS[, , "co2cement_process"]*(1-p_share_CCS)
+                  # fixme: missing/wrong emissions
+                  ) * GtC_2_MtCO2,
                  "Emi|CO2|+|Energy (Mt CO2/yr)"))
 
   ### 2.2 Non-energy CO2 emissions ----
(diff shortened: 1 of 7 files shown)


fbenke-pik:
The test also fails for ECEMF. The problem is the variable `emi.vars.wNonEn`, which is declared under the condition you introduced, is also used later on in line 2770: https://github.com/mellamoSimon/remind2/blob/fix-emissions-inconsistency/R/reportEmi.R#L2770

So not knowing anything about the specifics of this reporting, I can at least tell you that you would have to make sure that  `emi.vars.wNonEn` is defined with reasonable variables under any condition.

mellamoSimon:
ok, will do, thank you! For the record: I didn't realize that the mapping tests can be skipped and buildLibrary still is successful when running locally (which is what had happened here)

R/reportEmi.R
```diff
@@ -496,19 +494,20 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, t = c(seq(200
...
 
  # CO2 emissions from the end-of-life of carbon-bearing products
+ # fixme: wild guess: mybe I have to discount the carbon of biogenic and atmospheric origin from here as well. Nope, wasn't that
 if (!is.null(vm_plasticsCarbon)) {
     out <- mbind(out,
-               setNames(dimSums(vm_feedstockEmiUnknownFate, dim=3)* GtC_2_MtCO2,
+               setNames((dimSums(vm_feedstockEmiUnknownFate, dim=3))* GtC_2_MtCO2,
                         "Emi|CO2|Energy|Waste|+|Feedstocks unknown fate (Mt CO2/yr)"),
-               setNames(dimSums(vm_incinerationEmi, dim=3)* GtC_2_MtCO2,
+               setNames((dimSums(vm_incinerationEmi, dim=3))* GtC_2_MtCO2,
```
0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, line 510: What is up with those brackets?

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

mellamoSimon:
I ran all the mapping tests manually and they all passed (with warnings of missing variables for some and the gaps in the summation checks, as expected)

R/reportEmi.R
mellamoSimon: ups, I was trying something and forgot to delete those. Now I did.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
:shipit:

Issue #524, opened 2024-02-06
Drop Backwards Compatibility


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
**tl;dr**
- not necessary anymore, as projects can use `renv` to manage their R package
  environment
- creates Byzantine code nobody is able to follow
- causes extremely long test times (half an hour) for questionable tests
- in summary a huge time sink
- check for the existence and plausibility of variables and values for a core
  set of module realisations instead

Backwards compatibility was introduced into the `remind2` package (and before
that in the `remind` package) when we did not have a way to properly manage R
package versions on the cluster, and everybody and every project used the common
library of R packages which was (is?) updated whenever a new version of a RSE
package is merged on Github.

But now projects can easily manage their R package environments and keep it
stable with whatever package versions they need and that works for them.
Updates to the `remind2` package are usually tied to updates in the REMIND code.
Either projects track the REMIND `develop` branch to benefit from the REMIND
updates — then they will want (need) to get new `remind2` versions, too.  Or
they do not track the REMIND developments, in which case there is no point for
them to update the `remind2` package, either.

Currently, we are actively tracking compatibility with five projects

- NGFS #1487 (7 December 2023) but closed
- Ariadne #1430 (16 October 2023) +11 commits
- SHAPE #1303 (21 April 2023) but closed
- ECEMF #1268 (29 March 2023)
- NAVIGATE #982 (27 September 2022) but closed

all of which closely tracked REMIND development and did not diverge far from it.
So there is no benefit for them from updated `remind2` package versions, either.

All the backwards compatibility leads to a Byzantine code structure that nobody
can follow anymore, resulting in increasing complexity and errors, as well as
excessive test times (around 30 minutes currently).

I suggest to ditch the backwards compatibility, and focus on a correctly working
post-processing for current REMIND code instead.  Define a set of module
realisations that have to be supported, generate up-to-date gdxes for them as
part of the AMTs, and test for the existence and plausibility of variables and
values in the output.

fbenke-pik:
I am not sure I fully understand what exactly you want to get rid of, Michaja. But in general, I agree that there is potential. 

Just a few quick thoughts that come to my mind:
- I personally am not a fan of the conditional reporting as it is currently done for Ariadne in reportEmi. I think it was supposed to be a temporary hotfix, but it has been in the code forever and is causing problems. Can we think of a better way to achieve the same? Or drop it altogether?
[...]

PR #534, opened 2024-02-15
fix and simplify NGFS+SHAPE piamInterfaces test

orichters:
- Noticed that [this code](https://github.com/pik-piam/remind2/blob/61607b5e07c6e847ba9e16aed6bf46a1a72f2a20/tests/testthat/test-piamInterfaces-NGFS.R#L19-L22) did not actually work, because `unique("a", "b")` is just `"a"`, so a `c()` was missing there. So actually NGFS and SHAPE were never tested, but only the `AR6` part.
- Instead of running own tests on NGFS and SHAPE gdx files, integrate the piamInterfaces check into the usual convGDX2MIF test
- Add a check for the latest NPi AMT run
- cleanup as part of https://github.com/pik-piam/remind2/issues/524


tests/testthat/test-convGDX2mif.R
```diff
@@ -28,36 +28,56 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
...
   # finds for each AMT scenario the most recent, successfully converged GDX
-  .findAMTgdx <- function(gdxPaths = NULL) {
+  .findAMTgdx <- function(gdxPaths = NULL, scenario = NULL) {
     didremindfinish <- function(fulldatapath) {
       logpath <- paste0(stringr::str_sub(fulldatapath, 1, -14), "/full.log")
       return(file.exists(logpath) &&
         any(grep("*** Status: Normal completion", readLines(logpath, warn = FALSE), fixed = TRUE)))
     }
-    gdx <- Sys.glob("/p/projects/remind/modeltests/output/*/fulldata.gdx")
+    gdx <- Sys.glob(paste0("/p/projects/remind/modeltests/remind/output/", scenario, "*/fulldata.gdx"))
```
fbenke-pik, line 37: this expects you are on the cluster, right?

orichters: Yes, the AMT test is skipped if you are not on the cluster.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):
Does anybody ever run tests on the cluster?  Does anybody ever uncomment/amend stuff inside the tests when running on the cluster?  Or is that all just dead code?

It took me a while to parse `.findAMTgdx()`, so I refactored it. :point_right: https://github.com/orichters/remind2/pull/2

PR #538, opened 2024-02-16
add test for ranges on convGDX2MIF() output

### LINKED ISSUE #526: Add structure for range checks
Should be possible to test stuff like `Emi|CO2|Energy|Demand|Industry|Chemicals|Solids|+|Fossil > 0`.

FYI @fbenke-pik.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
close #526


pweigmann (approved):

fbenke-pik (approved):

orichters:
Wouldn't it make sense to run this check every time `convGDX2MIF` is run, not only in the tests? If there is such an inconsistency in any run, it would be good to know…

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
That dawned on me over the weekend as well.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
I will hold this back anyhow, since the test would currently break `remind2` building (because `remind2` is broken).

orichters:
Or, you could: turn `stop` into `warning`, accept that warning in `.buildLibrary`, and merge it, so people working on the issues can more easily keep track of these inconsistencies (and make sure we don't add some).

Of course, this implies that people would actually care about these test warnings. History shows they don't.

pweigmann (approved):
Definitely looks very similar to a special application case of the validation tool now, would be curious to see how performance compares. Support for regex variable selection is the only thing I can see right now that is not available in [``validateScenarios``](https://github.com/pik-piam/piamValidation/blob/main/R/validateScenarios.R).

PR #543, opened 2024-02-19
The negation seems to have survived unintentionally

dklein-pik:


orichters (approved):
🤦‍♂️ 

Thanks for fixing it!

PR #544, opened 2024-02-19
include isVersionUpdated check

fbenke-pik:
This PR activates checking if buildLibrary was run in this package when creating a PR


fbenke-pik:
in https://github.com/pik-piam/remind2/blob/master/.github/workflows/check.yaml

fbenke-pik:
It is disseminated via lucode and was added in a previous PR (David was faster)

LaviniaBaumstark:
ah, was woundering why it was not part of your PR :-)

LaviniaBaumstark (approved):

PR #546, opened 2024-02-22
adjust skip condition in unit tests

fbenke-pik:


fbenke-pik:
Dropped, unless we find a way to make `skip_on_ci` with R-Universe

Issue #552, opened 2024-02-29
Summation errors due to CDR-related energy


katarkow:
Due to the assumption that emissions from using `fegas` to produce heat for DAC are captured, there are summation errors in the reporting. It happens because  `vm_emiTeMkt` accounts for burning all that `fegas` used for DAC and also synfuels from that captured carbon, so we account for that carbon twice. I thought of the following solutions two solutions:

(1) Updating the `vm_emiTeMkt` in REMIND, so discounting these emissions * share CCU or
(2) Discounting the CCU from gas for DAC in remind2, (a) either from `vm_emiTeMkt` and `vm_emiAllMkt` directly where they’re declared or (b) everywhere where they’re used, so `Emi|CO2`, `Emi|CO2|+|Energy`, `Emi|GHG`, …. (potentially painful to remember about this when a new reporting variable uses either of the REMIND variables).

What are your thoughts?

I tried (1)  [here](https://github.com/remindmodel/remind/pull/1589), I thought it would be the cleanest. However, there's a small accounting error in the reporting from using `sm_eps` to handle cases in which there’s no captured CO2 in a given (t, regi): `share = ccs / (captured + sm_eps)`.

katarkow:
I tried (1)  [here](https://github.com/remindmodel/remind/pull/1589), I thought it would be the cleanest. However, there's a small accounting error in the reporting from using `sm_eps` to handle cases in which there’s no captured CO2 in a given (t, regi): `share = ccs / (captured + sm_eps)`. Anybody has a better idea to handle this? :)

(2a) Is here: #553

LaviniaBaumstark:
is this still an issue in latest runs: /p/projects/remind/runs/REMIND_2024_09_24/remind/output ?

orichters:
`/p/projects/remind/runs/REMIND_2024_09_24/remind/output/SSP2-PkBudg650_2024-09-24_13.59.41/log.txt` still has this error:
```
Carbon Sequestration|CCS >                                                    + Carbon Management|Storage >
   + Carbon Sequestration|CCS|Biomass                                            + Carbon Management|Storage|+|Biomass|Pe2Se + Carbon Management|Storage|Industry Energy|+|Biomass
   + Carbon Sequestration|CCS|Fossil                                             + Carbon Management|Storage|+|Fossil|Pe2Se + Carbon Management|Storage|Industry Energy|+|Fossil
[...]

Issue #445, opened 2023-08-24
Emi|GHG|ESR and Emi|GHG|++|Outside ETS and ESR reporting mismatch


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
The newly implemented [summation checks](https://github.com/pik-piam/remind2/pull/444) show a mismatch between `Emi|GHG|++|ESR` and its `Emi|GHG|ESR|+|x` components, as well as `Emi|GHG|++|Outside ETS and ESR` and its `Emi|GHG|Outside ETS and ESR|+|x` components for the `indu_subsectors-buil_simple-tran_edge_esm-POP_pop_SSP2EU-GDP_gdp_SSP2EU-En_gdp_SSP2EU-Kap_debt_limit-Reg_62eff8f7.gdx` (probably others, too).
```
Emi|GHG|++|ESR <                                                               <
   + Emi|GHG|ESR|+|Industry                                                      + 
   + Emi|GHG|ESR|+|Transport                                                     + 
   + Emi|GHG|ESR|+|Buildings                                                     + 
   + Emi|GHG|ESR|+|Agriculture                                                   + 
   + Emi|GHG|ESR|+|Waste                                                         + 
Relative difference between -1.4% and -1%, absolute difference up to 11.29 Mt CO2eq/yr.
```
```
Emi|GHG|++|Outside ETS and ESR >                                               >
   + Emi|GHG|Outside ETS and ESR|+|Transport                                     + 
   + Emi|GHG|Outside ETS and ESR|+|Land-Use Change                               + 
   + Emi|GHG|Outside ETS and ESR|+|F-Gases                                       + 
Relative difference between -186.4% and 48.2%, absolute difference up to 29.38 Mt CO2eq/yr.
```

orichters:
Note: Because of [this error](https://github.com/pik-piam/piamInterfaces/issues/158), it need not be the case that this is the actual error. Let us maybe sort that out first.

orichters:
After fixing the error in checkSummations, the error persists (using `indu_subsectors-buil_simple-tran_edge_esm-POP_pop_SSP2EU-GDP_gdp_SSP2EU-En_gdp_SSP2EU-Kap_debt_limit-Reg_62eff8f7.gdx`):
```
Emi|GHG|++|ESR <>
   + Emi|GHG|ESR|+|Industry
   + Emi|GHG|ESR|+|Transport
   + Emi|GHG|ESR|+|Buildings
[...]

PR #553, opened 2024-02-29
Fix DAC-related emissions

katarkow:
Captured emissions from using natural gas for DAC’s heat are accounted for twice: burning natural gas and released when using synfuels. It could be fixed by discounting them from `vm_emiTeMkt` and `vm_emiAllMkt` in the reporting. See #552.

The remaining summation errors are not CDR-related:
```
Emi|GHG|++|ESR <>                                                        	 
   + Emi|GHG|ESR|+|Industry                                           	 
   + Emi|GHG|ESR|+|Transport                                          	 
   + Emi|GHG|ESR|+|Buildings                                          	 
   + Emi|GHG|ESR|+|Agriculture                                        	 
   + Emi|GHG|ESR|+|Energy Waste                                       	 
   + Emi|GHG|ESR|+|Waste                                              	 
Relative difference between -1.4% and 0.0361%, absolute difference up to 31.8 Mt CO2eq/yr.

Emi|GHG|++|Outside ETS and ESR <>                                        	 
   + Emi|GHG|Outside ETS and ESR|+|Transport                          	 
   + Emi|GHG|Outside ETS and ESR|+|Land-Use Change                    	 
   + Emi|GHG|Outside ETS and ESR|+|F-Gases                            	 
Relative difference between -15519% and 14.8%, absolute difference up to 31.8 Mt CO2eq/yr.
```

### DIFF
--- a/R/reportEmi.R
+++ b/R/reportEmi.R
@@ -1396,7 +1396,7 @@ if (!is.null(vm_plasticsCarbon)) {
                setNames(dimSums(vm_emiIndCCS[, , "co2cement_process"], dim = 3) * GtC_2_MtCO2,
                           "Carbon Management|Carbon Capture|Industry Process|+|Cement (Mt CO2/yr)"),
                # total co2 captured by DAC
-               setNames(-vm_emiCdrTeDetail[, , "dac"] * GtC_2_MtCO2,
+               setNames(dimSums(vm_ccs_cdr, na.rm = TRUE) * GtC_2_MtCO2,
                           "Carbon Management|Carbon Capture|+|DAC (Mt CO2/yr)"),
                # total co2 captured
                setNames(vm_co2capture * GtC_2_MtCO2,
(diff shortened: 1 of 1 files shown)


orichters:
👍 

> The remaining summation errors are not CDR-related:

And quite old: https://github.com/pik-piam/remind2/issues/445

fschreyer (changes requested):
Thanks. Good to see that it adds up. Sry, I did not see that before and commented here #552. However, I would not subtract the gas capture from the ``vm_emi...`` variables when they are read in as this will get even more confusing for people in the future trying to match this with REMIND code. 

I think it would be better just to add ``co2capture_DAC_energy`` to [this line](https://github.com/pik-piam/remind2/blob/8dd70fd28ee5d1d97705193d148040d314db93e1/R/reportEmi.R#L1178) and [this line](https://github.com/pik-piam/remind2/blob/8dd70fd28ee5d1d97705193d148040d314db93e1/R/reportEmi.R#L1127) and add a short comment above. 

Moreover, I think it needs some change in the REMIND code as well, see #552.

PR #554, opened 2024-03-01
Fix carbon price reporting for EU21 runs if sectorally differentiated carbon taxes via switch cm_CO2TaxSectorMarkup are used

fschreyer:
For running REMIND-EU with cm_CO2TaxSectorMarkup, the reporting needed some fixing.

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '224621680'
+ValidationKey: '224675531'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,8 +2,8 @@ cff-version: 1.2.0
 message: If you use this software, please cite it using the metadata from this file.
 type: software
 title: 'remind2: The REMIND R package (2nd generation)'
-version: 1.135.6
-date-released: '2024-02-27'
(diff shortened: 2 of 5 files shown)


orichters (approved):
if you say so :)

Don't really get what the problem was.

But thanks for fixing!

fschreyer:
> Don't really get what the problem was.
> 

I could have been more explicit. Essentially, the problem was that ``output_wo_GLO`` comes with all aggregated regions like ``EU27`` etc., while ``p21_CO2TaxSectorMarkup`` only comes with REMIND regions. So, I guess the lesson is that one needs to mind that when combining the output objects of reporting functions, which feature more regions, with parameters or variables that get read in from REMIND directly.

orichters (approved):

orichters:
@fschreyer: Another thought: Did you see [this warning](https://github.com/fschreyer/remind2/blob/aff4ecc6cecfe140309adf0abe78c74bfc4a9f72/R/reportPrices.R#L752)? I think using sectorally differentiated carbon taxes with EU21 is not advisable, as of now.

fschreyer:
> @fschreyer: Another thought: Did you see [this warning](https://github.com/fschreyer/remind2/blob/aff4ecc6cecfe140309adf0abe78c74bfc4a9f72/R/reportPrices.R#L752)? I think using sectorally differentiated carbon taxes with EU21 is not advisable, as of now.

Thanks. I did not see it.

PR #560, opened 2024-03-11
remove FE|Industry|Chemicals|Heat from reporting, which is always zero anyhow

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


orichters (approved):
Thanks

Issue #549, opened 2024-02-27
memory and mbind problems in convGDX2MIF / reportFE


orichters:
- Running output.R -> single -> reporting.R runs out of memory with standard --mem=8000 setting, even --mem=12000 fails.
- Reason seems to be that reportFE requires more memory
- Using output.R -> single -> reporting, @orichters started output generation for several AMT testOneRegi runs. Result: Until 2023-12-22, it works, afterwards, it fails, see `cd /p/projects/remind/modeltests/remind/output; grep oom-kill testOne*/log_output.txt`
- @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q found out it is gdx::readGDX(), triggered by vm_demFENonEnergySector:
```
> foo <- gdx::readGDX('./output/testOneRegi-Base-AMT_2024-02-16_22.08.20/fulldata.gdx', 'vm_demFENonEnergySector', field = 'l', restore_zeros = TRUE, react = 'silent')
> format(object.size(foo), units = 'auto')
[1] "3 Gb"
> bar <- gdx::readGDX('./output/testOneRegi-Base-AMT_2024-02-16_22.08.20/fulldata.gdx', 'vm_demFENonEnergySector', field = 'l', restore_zeros = FALSE, react = 'silent')
> format(object.size(bar), units = 'auto')
[1] "138.9 Kb"
```
- According to @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, one cannot simply set `restore_zeros = FALSE` because we declare all variables over the general supersets (e.g., `tall`, `all_enty`), but usually define them only over subsets (`t`, `entyFE`). `restore_zeros = FALSE` reads in only data that is not zero, which is usually what we want. But if some axis of the data (e.g. all values in 2010, or everything `fehos`) happens to be entirely zero (usually that happens for specific gdxes only), that dimension of the resulting magpie object differs (e.g. c(2005, 2015, …) instead of c(2005, 2010, 2015, …)) and it is incompatible with all the rest of the objects we have. What would be needed is a function to expands a variable read in with `restore_zeros = FALSE` to the actual sets it should have. But that is a moving target, as it depends on all the other variables/parameters used in conjunction with it.
- `magpie_expand` is [not the solution](https://github.com/pik-piam/magclass/issues/166)

Some code to start with:
```
library(magclass)
t <- c(1995, 2005, 2015)
s <- c('AFR', 'CPA')
[...]

PR #563, opened 2024-03-13
add expandMagclass helper

fbenke-pik:
Adds a helper to reshape read in magclass objects to match a reference so subsequent operations do not create problems due to mismatched dimensions.

See also https://github.com/pik-piam/remind2/issues/549


R/reportEmi.R
```diff
@@ -143,8 +143,11 @@ reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL,
   # FE non-energy use
   vm_demFENonEnergySector <- readGDX(gdx, "vm_demFENonEnergySector", field = "l",
                                      restore_zeros = F, react = "silent")[,t,]
+
   if (length(vm_demFENonEnergySector) == 0) {
     vm_demFENonEnergySector <- NULL
+  } else {
+    vm_demFENonEnergySector <- expandMagclass(vm_demFENonEnergySector, vm_demFeSector)
```
fbenke-pik, line 150: as this helper also removes values from third dimension that are not in `vm_demFeSector`, i am not sure if the trimming should be done here or only further down below in the context of the condition where it crashes (line 304)

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q: Here is good.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

orichters (approved):
thanks (probably you are still running buildLibrary?)

Maybe the proper place for this function would be in `magclass` somewhere, but I actually don't care.

fbenke-pik:
> thanks (probably you are still running buildLibrary?)

I will, it just takes forever for remind2

PR #579, opened 2024-03-27
add |+| to 'Energy Investments', remove DAC from Electricity

orichters:
- Energy Investments had a summation issue: https://github.com/remindmodel/development_issues/issues/265
- Turns out: [this line](https://github.com/pik-piam/remind2/blob/1886a00d5fec0dc7f028b7cecad41a5e6188d3bc/R/reportEnergyInvestment.R#L171) calculating `Energy Investments|Electricity` also sums over `dac` which is part of `tenotransform` which is [added here](https://github.com/remindmodel/remind/blob/d0596c7aad076387808ff75cf0a14271417d67d7/modules/33_CDR/portfolio/sets.gms#L70), while this is no sub-category of `Energy Investments|Elec|*`. So I removed the CCS technologies from the Electricity summation
- At some later point, the `Energy Investments|Demand|*` chain might be populated with (not only) the CDR data
- Added `|+|` notation and additional summation checks
- In renamed `Energy Investments|Elec` to `Energy Investments|Electricity` like all the parent category. This will need some fixes [to the piamInterfaces mappings](https://github.com/pik-piam/piamInterfaces/pull/290/files) to keep backwards compatibility. Also did that for `Operational costs|Elec|wind` and `Total Energy costs|Elec|Biomass`
- One issue: the old gdx from the last REMIND release fails on one of them, but the newest AMTs don't. Probably, there was something fixed in REMIND in between.
- in cs2, rename all incoming `|Elec|` to `|Electricity|` to allow for backwards compatibility

See
```
devtools::load_all("/p/tmp/oliverr/piamInterfaces")
devtools::load_all("/p/tmp/oliverr/remind2")
gdx <- "/p/projects/remind/runs/REMIND_2024_03_24/remind/output/SSP2EU-PkBudg650_2024-03-24_20.02.18/fulldata.gdx"
d <- checkSummations(reportEnergyInvestment(gdx), outputDirectory = NULL, summationsFile = "extractVariableGroups")
d <- checkSummations(reportEnergyInvestment(gdx), outputDirectory = NULL, summationsFile = "/p/tmp/oliverr/remind2/inst/extdata/additional_summation_checks.csv")
```


inst/markdown/compareScenarios2/cs2_main.Rmd
```diff
@@ -238,6 +238,9 @@ data <-
         unit = "billion US$2005/yr",
         value = value / 1000)
   )
+# backwards compatibility for https://github.com/pik-piam/remind2/pull/579
+data <- data %>%
```
fbenke-pik, line 242: Why are you changing back the renaming here?

fbenke-pik: Ah, now I understand. Once again, I question this approach to backwards compatibility. If you have old runs, why not use the fitting remind2? These kind of backwards compatibility adjustments come with technical debt and clutter the code. So if you are certain, it has a real life use, keep it. Otherwise, drop it.

orichters: I'm not changing back. If I get old data that still has `|Elec|` variables (for example all the AMT comparisons with older runs, or comparing an old release version to today's runs), I convert them into `|Electricity|` such that they can be plotted.

fbenke-pik (approved):

PR #581, opened 2024-04-02
set spatial region when reading vm_demFENonEnergySector from gdx

fbenke-pik:


fschreyer (approved):
Great, thanks for the fast fixing.

PR #583, opened 2024-04-05
Download AMT gdx for test of convGDX2MIF

dklein-pik:


tests/testthat/test-convGDX2mif.R
```diff
@@ -16,48 +16,18 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
...
-
-      # duplicates are older than the preceding runs
-      dup <- duplicated(sub(paste0("_", datetimepattern), "",
-                            basename(dirname(path))))
+  gdxList <- c("fulldata-release.gdx" = "https://rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_fulldata.gdx",
+               "fulldata-AMT.gdx"     = "https://rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_SSP2EU-NPi-AMT.gdx")
 
-      # everything that is no duplicate is the latest run of that scenario
-      return(path[!dup])
+  if (length(gdxPaths) == 0) {
```
dklein-pik, line 22: Do we need still this check?

tests/testthat/test-convGDX2mif.R
```diff
@@ -16,48 +16,18 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
...
-        defaultGdxPath,
-        mode = "wb", quiet = TRUE
-      )
-    }
-    gdxPaths <- defaultGdxPath
-  }
-
-  # finds for each AMT scenario the most recent, successfully converged GDX,
-  # that is no older than 30 days
-  .findAMTgdx <- function(gdxPaths = NULL, scenario = NULL) {
```
dklein-pik, line 32: Do you agree that we dont need this functio anmore, because the AMT gdxes will be uploaded to the RSE server by the AMT script?

tests/testthat/test-convGDX2mif.R
orichters: No, I think you should always run the checks of the `gdxList`.

tests/testthat/test-convGDX2mif.R
```diff
@@ -16,48 +16,18 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
...
+  gdxList <- c("fulldata-release.gdx" = "https://rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_fulldata.gdx",
+               "fulldata-AMT.gdx"     = "https://rse.pik-potsdam.de/data/example/remind2_test-convGDX2MIF_SSP2EU-NPi-AMT.gdx")
 
-      # everything that is no duplicate is the latest run of that scenario
-      return(path[!dup])
+  if (length(gdxPaths) == 0) {
+    for (i in seq_along(gdxList)) {
+      from <- gdxList[i]
+      to   <- file.path(tempdir(), names(gdxList[i]))
+      if (!file.exists(to)) {
```
orichters, line 26: Not sure whether that makes sense. Because once you have a gdx file there, you will never download the new AMT file, right? Can you maybe check with @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q on how you can integrate the caching he suggested [here](https://github.com/pik-piam/remind2/pull/478/files) (but somehow never merged)?

tests/testthat/test-convGDX2mif.R
orichters: Fine for me. I used it sometimes, but I guess nobody else did. And as running `make build` on the cluster does not really work anyway at the moment, you can delete it.

tests/testthat/test-convGDX2mif.R
```diff
@@ -83,8 +53,6 @@ test_that("Test if REMIND reporting is produced as it should and check data inte
   # gdxPaths <- c(gdxPaths, Sys.glob("/p/projects/remind/inputdata/CESparametersAndGDX/*.gdx"))
   # uncomment to add debugging example gdx files
   # gdxPaths <- c(gdxPaths, Sys.glob("/p/projects/remind/debugging/gdx-examples/*.gdx"))
```
orichters, line 55:
```suggestion
```
I think you can delete that as well.

orichters (approved):

tests/testthat/test-convGDX2mif.R
dklein-pik: Interesting. Isn't the tempdir() different every time it is created?

orichters: Ah, yes, you are right for course.

dklein-pik: Which means the gdx files will be downloaded again and again, which Michaja wanted to avoid if I unterstand correctly. I'll merge anyway and we can still integrate Michaja's solution.

PR #587, opened 2024-04-15
Delete non-ldv entries (extensive)

jmuessel:
In the new version, using [reporttransport](https://github.com/pik-piam/reporttransport), non-ldv variables will not be reported anymore. With this PR 
1. non-ldv variables that are related to COMPLEX will be deleted. 
2. For this to function the required minimal version of connected packages is stated. 

In another PR, I already renamed variables and deleted non-LDV variables from rmd skripts but not from the other R skripts as I am not sure about the consequences - see discussion below.

ToDos 
- build package 
- check that in other packages the correct remind2 version is required (PIAM interfaces e.g.)


R/reportPrices.R
```diff
@@ -342,7 +342,6 @@ reportPrices <- function(gdx, output=NULL, regionSubsetList=NULL,
       fe = c(
         fehos = "Liquids",
         fepet = "LDV|Liquids",
```
fbenke-pik, line 344: @orichters Due to deletion of 'fedie' , the function `addVar` throws the warning `In reportPrices, addVar called with a NA value, e.g. "Price|Final Energy|Transport|other|NA|Fossil (US$2005/GJ)". 

What would be the correct way to deal with this? Not delete the `fedie` entries or somehow address them being removed?

fbenke-pik: Btw, I think a warning in in addVar is not sufficient when there are NAs. Should crash or return NULL, as the returned magclass object becomes unusable, yet it is not easy to find the source of the problem.

orichters: Please consult with @Renato-Rodrigues who [introduced the addVar function](https://github.com/pik-piam/remind2/commit/24f46abbdfab7e55d3b0bf07bb5e2ca5870cbbba#diff-9079799de4d750220602928460d4d1998c56d82cbb46b1f6b1c95405768aec36R374). I don't understand how this can create duplicates, though: [this line](https://github.com/pik-piam/remind2/blob/7927dad0dfae6cb860985aa745e02148db23a53f/R/reportPrices.R#L387-L389) should avoid it.

fbenke-pik: You are right, the duplicates are produced in another part of the code. 

We still have to avoid the warning, @jmuessel  as it will be thrown 50+ times, i.e. individual warnings are no longer printed and do not appear in the log.

fbenke-pik:
Closing, due to no activity.

PR #589, opened 2024-04-17
move deletePlus to piamUtils package

fbenke-pik:


R/reexport.R
```diff
@@ -0,0 +1,8 @@
+#' @importFrom piamutils substitutePlaceholder
+#' @export
```
orichters, line 2: not sure we need the substitutePlaceholder function, because its name was substituePlaceholder with a typo anyway and I think nobody used it outside remind2.

orichters (approved):

R/reexport.R
fbenke-pik: correct, i will remove it

orichters (approved):

PR #590, opened 2024-04-18
fix name of helper in cs2_main

fbenke-pik:


orichters (approved):
did you load that from piamUtils?

fbenke-pik:
> did you load that from piamUtils?

Yes, I made this explicit now

PR #594, opened 2024-05-03
Report climate assessment variables

tonnrueter:
- Function  `reportClimate` extracts variables `pm_globalMeanTemperature` & `p15_forc_magicc` from `fulldata.gdx`
- Adapted `convGDX2MIF.R` to include climate assessment variables


R/reportClimate.R
```diff
@@ -0,0 +1,37 @@
...
+#' @examples
+#' \dontrun{
+#' reportClimate(gdx)
+#' }
+#'
+#' @export
+#' @importFrom gdx readGDX
+#' @importFrom magclass setNames mbind
+reportClimate <- function(gdx) {
+  tryCatch(
```
fbenke-pik, line 17: tryCatch can most likely be avoided here. `readGDX` has a `react` parameter  that can be set to `silent` to return NULL

R/reportClimate.R
```diff
@@ -0,0 +1,37 @@
...
+    # Climate assessment is turned of per default. Try reading climate asessment data from GDX, but fail
+    # silently if none are available.
+    {
+      variables <- c(
+        pm_globalMeanTemperature = "Temperature|Global Mean (K)",
+        p15_forc_magicc = "Forcing (W/m2)"
+      )
+      return(mbind(lapply(names(variables), function(gamsVariable) {
```
fbenke-pik, line 25: I am not a fan of lapply here. Especially, if you are looping only over two variables. These lines are hard to read, understand and debug. You should simply write two readGDX statements for the sake of clarity.

R/convGDX2MIF.R
```diff
@@ -126,6 +126,30 @@ convGDX2MIF <- function(gdx, gdx_ref = NULL, file = NULL, scenario = "default",
     message("function reportSDPVariables does not work and is skipped")
   }
 
+  # Report climate assessment variables & merge with output
+  message("running reportClimate...")
+  climateVars <- reportClimate(gdx) # Returns NULL if no climate assessment variables are available
+  if (!is.null(climateVars)) {
```
fbenke-pik, line 132: All this should be moved to `reportClimate`. convGDX2MIF is the hub function and should only have two lines of code, sth along the lines of

```
message("running reportClimate...")
output <- mbind(output,reportClimate(gdx, ...))
```

R/reportClimate.R
tonnrueter: Done

R/convGDX2MIF.R
tonnrueter: Moved to `reportClimate.R`

fbenke-pik (approved):

PR #600, opened 2024-05-17
Define gdx files that are donwloaded and used for tests

dklein-pik:
Three dimensions need to be covered when selecting the gdx files for the test with convGDX2MIF: **regional resolution** (H12 and EU21), **REMIND version** (release and AMT), and **climate policy** (NPi and PkBudg). To reduce the number of gdx files that are tested with convGDX2mif during buildLibrary(), we combine them as follows: 

- test fulldata.gdx from SSP2EU-**EU21-NPi**_2024-03-28_10.23.13 from the latest release, and 
- test fulldata.gdx from **SSP2EU-PkBudg650**-AMT from the latest AMT.

Further reduce number of gdx files tested by removing the individual tests `test-piaminterfaces-Ariadne.R`, `test-piaminterfaces-ECEMF.R`, and `test-piaminterfaces-NAVIGATE.R`. However, add Ariadne and ECEMF to the list of templates that are checked with checkPiamTemplates. NAVIGATE is outdated.


dklein-pik:
> be aware that this will need to be adjusted soon when moving from SSP2EU to SSP2

I'll try. And then we also need to upload a different gdx from the AMTs.

orichters (commented):
Thanks, approve it :)

PR #605, opened 2024-05-27
Add capacity and investment variables for DAC

fschreyer:
This was demanded by the ariadne project. 

Adds the variables

```
Cap|Carbon Management|DAC
New Cap|Carbon Management|DAC
Cumulative Cap|Carbon Management|DAC
Energy Investments|DAC.
```

Example Mif is here: ``/p/tmp/schreyer/Modeling/remind/FossilFree/output/EU27_NZ_CCS100Mt_2024-03-11_15.21.21/test_DAC_cap.mif``.


R/reportCapacity.R
tabeado, line 1: 1. Is cap in native REMIND units? the native dac unit is GtC, so if you just multiply by 3.66 the unit would be Gt CO2/yr.
2.  Do we have a standardized conversion factor for C to CO2 in remind2? In REMIND it is 3.666666666667, I would have used 3.67. Not a big difference of course, but might lead to checks not adding up?

tabeado (changes requested):

R/reportCapacity.R
fschreyer: Thanks. 

On 1.) ``vm_cap`` is already multiplied by 1000 before (see [here](https://github.com/pik-piam/remind2/blob/eba912d5dabace5f5632ea77a76b45e30f8f5088/R/reportCapacity.R#L39)). ;-)

On 2.) Should be addressed now.

tabeado (approved):

PR #606, opened 2024-06-03
remove code of old 11_aerosols realizations

orichters:
- remove reporting code for the postIIASA and postECLIPSE realizations of the 11_aerosols module that maybe have existed in ancient times before we moved to github in 2019.
- The only thing I have done is remove everything that is behind the `if` switches with the two old realizations. Unfortunately, the full diff is basically unreadable. Check commits separately. The `adjust whitespace` is really just removing two blanks, but the diff looks terrible.
- https://mattermost.pik-potsdam.de/rd3/pl/6jy76asjsif5dg6a8q8ei4u9bh
- in line with https://github.com/pik-piam/remind2/issues/524


gabriel-abrahao (approved):

gabriel-abrahao:
Looks good to me, and I also can't see any use for that old code now.

PR #607, opened 2024-06-11
Add basicmode to reportEmi

gabriel-abrahao:
The new MAGICC7_AR6 climate implementation requires nicely formatted emissions data.

`reportEmi` currently needs a lot of sector-specific information to provide the breakdown of CO2 emissions. In REMIND, module `15_climate` runs before some of the modules that generate this information, making it not work on a GDX that was generated in the postsolve in the first iteration. 

This PR adds a `basicmode` optional argument to `reportEmi`, that circumvents the need for sector-specific information by calculating `Energy and Industrial Processes` CO2 emissions via the difference between total and LUC emissions, and reports only the emissions variables needed for the climate assessment. Since the function is a bit messy at the time, it's hard to detach these from the current code, so there's quite a bit of code repeating here. However, the only out of this might be a complete refactor of `reportEmi`. Putting this functionality in a separate function might be worse than this solution, as it wouldn't be maintained as closely.

### DIFF
--- a/R/reportEmi.R
+++ b/R/reportEmi.R
@@ -25,11 +25,11 @@
 #' @importFrom tibble as_tibble
 #' @importFrom piamutils deletePlus
 
-reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL,
+reportEmi <- function(gdx, output = NULL, regionSubsetList = NULL, basicmode = FALSE,
                       t = c(seq(2005, 2060, 5), seq(2070, 2110, 10), 2130, 2150)) {
 
   # emissions calculation requires information from other reporting functions
-  if (is.null(output)) {
+  if (is.null(output) & !basicmode) {
     message("reportEmi executes reportFE")
     output <- mbind(output, reportFE(gdx, regionSubsetList = regionSubsetList, t = t))
   }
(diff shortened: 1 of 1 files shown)


gabriel-abrahao:
@tonnrueter `buildLibrary()` is complaining about a million linter problems, including many things that definitely were there before. Would you mind taking a look?

pweigmann (approved):
I don't see a general problem with this approach, but it might be good to get the opinion of @fschreyer on this as well.

tonnrueter:
In its current state adjustments to `reportEmi` do not seem advisable. Therefore, Gabriel & I agreed to instead implement a dedicated `reportEmiForClimateAssessment` function that extracts the quantities needed for climate assessment in between GAMS iteration from `fulldata_postsolve.gdx`. Although we risk to miss bug fixes etc to `reportEmi`, this seems the most practicable way to go foward.

fschreyer:
The emissions equations in the REMIND core and the emissions reporting in reportEmi are unfortunately quite complex and not ideally structured. We consider doing a general refactoring once the time allows it.

fbenke-pik:
@fschreyer the sooner the better. let me know if i can help somehow.

fschreyer:
> @fschreyer the sooner the better. let me know if i can help somehow.

Thanks, good to know. However, this will be a bigger process where we first need to get conceptually clear which kind of structure we want depending on current and future developments and need to coordinate across the group. I could lead this but I won't have time to start this process before spring 2025.

PR #609, opened 2024-06-18
correct reporting of industry carbon captured for chemical feedstocks

fschreyer:


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
`pm_IndstCO2Captured` is essentially Emissions × Capture Share
- Emissions is (FE - NE) × factor
- Capture Share is &sum; CCS / &sum; Emissions, so &sum; CCS / &sum; (FE - NE) × factor
so `pm_IndstCO2Captured` = (FE - NE) × factor × &sum; CCS / &sum; (FE - NE) × factor
_but_ NE happens to be just an exogenous share of FE so actually
`pm_IndstCO2Captured` = FE × &lambda; × factor × &sum; CCS / &sum; FE × &lambda; × factor &hArr; FE × &lambda; × factor × &sum; CCS / &lambda; × &sum; FE × factor &hArr; FE × factor × &sum; CCS /&sum; FE × factor
So _here and now_, non-energy demand actually does not matter.  Should still be fixed though, but mainly on the REMIND side, since this is just a quick fix for a REMIND bug that we cannot remove anymore, backwards compatibility be praised.

PR #612, opened 2024-06-20
Function reportEmiForClimateAssessment

gabriel-abrahao:
This PR adds a new function `reportEmiForClimateAssessment`. It functions as a simple version of `reportEmi` that can be run with an incomplete gdx in the first REMIND iterations, automatically runs and combines the output of `reportEmiAirPol`, and runs a lot faster. Only a few of the CO2 emissions variables are reported, with no sectoral detail, and just global numbers match those from `reportEmi`, as some bunker-related corrections are not done. 

The purpose is to be used in the `climate-assessment`/MAGICC7 related scripts that run between REMIND iterations.

It also temporarily ignores warnings related to an outdated GDX being used in `test-convGDX2mif`.


tonnrueter (approved):

.buildlibrary
```diff
@@ -1,11 +1,12 @@
-ValidationKey: '227436669'
+ValidationKey: '227587360'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
 - 'Warning: namespace ''.*'' is not available and has been replaced'
 - .*qpdf.* is needed for checks on size reduction of PDFs
 - .*following variables are expected in the piamInterfaces.*
 - Summation checks have revealed some gaps.*
 - .*to check if summation groups add up.*
+- .* Negative numbers in weight. Dangerous, was it really intended?.*
```
orichters, line 9: Which calculation returns this? Seems to be using `madrat::toolAggregate()`.

orichters (approved):
looks good from my side. Assuming you have tested this, I'm fine.

.buildlibrary
fbenke-pik: I will look into this separately and try to get rid of the warning again.I asked Gabriel to add it here so our checks don't hinder his own work. Should not block merging this.

fbenke-pik (approved):

PR #614, opened 2024-06-21
bring back more strict range check in convGDX2MIF

fbenke-pik:


0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q (approved):

orichters:
Sorry, @fbenke-pik and @0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q, but I think this really is not the right way to go.

The way it is implemented now just kills the reporting completely and interrupts running cascades, and the underlying error doesn't seem to be fixed in the current REMIND release, see [https://mattermost.pik-potsdam.de/rd3/pl/gyyig7y997g6fpxq7yjugte3go (@pvjeetze)

I think none of these range checks should actually stop the reporting before the mif file is written, that is nonsense. Really, warning should be absolutely sufficient (and would also stop buildLibrary, by the way).

fbenke-pik:
I guess your comment is more appropriate under this PR? https://github.com/pik-piam/remind2/pull/538

I am fine either way, so I'll leave the discussion to you two.

orichters:
No, PR #538 introduced it as a warning, while this turned it into a `stop`. I will prepare a fix.

fbenke-pik:
Yes, I fixed it in the wrong place. Thanks for spotting and fixing. 
The conflict remains though: warning vs stop in the test case?

PR #616, opened 2024-06-25
move test_ranges stop from convGDX2MIF back to test-convGDX2mif

orichters:
- to avoid REMIND reporting to fail
- I guess that was just an error that @gabriel-abrahao turned it from `stop` to `warning` in test-convGDX2mif and then @fbenke-pik back from `warning` to `stop` in a different file

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '227832100'
+ValidationKey: '227863449'
 AcceptedWarnings:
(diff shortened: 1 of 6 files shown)


fbenke-pik:
> I guess that was just an error that @gabriel-abrahao turned it from stop to warning in test-convGDX2mif and then @fbenke-pik back from warning to stop in a different file

Correct. Sorry, my bad.

fbenke-pik (approved):

PR #617, opened 2024-06-26
round absolute emission values to single grams to get rid of values different from zero introduced by precision errors

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


orichters (approved):
Looks good, thanks.

PR #618, opened 2024-06-26
adjust to deprecated read.gdx(factors = )

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:


orichters (approved):
Assuming you know what you are doing. Why was this parameter there in the first place and why is it not needed anymore? Two lines of explanation would be great for these PRs.

0UmfHxcvx5J7JoaOhFSs5mncnisTJJ6q:
That parameter was introduced to enable users to get back a data frame with strings instead of factors, which really are more trouble then they are worth.
Since some unknown update of `gdxrrw`, one never gets back factors, to the parameter is pointless in any case.

Issue #619, opened 2024-06-27
reportCapacity fails if DAC is switched off


orichters:
`/p/tmp/oliverr/NGFS_v5/remind-3p3p1/output/C_o_lowdem-rem-2/fulldata.gdx`

```
running reportCapacity...
Error in h(simpleError(msg, call)) :
  error in evaluating the argument 'object' in selecting a method for function 'setNames': subscript out of bounds ("dac")
Calls: convGDX2MIF ... setNames -> dimSums -> [ -> [ -> .local -> .dimextract
```

Probably caused by this change: https://github.com/pik-piam/remind2/commit/c6714e6db479a70a6d36ae6cb600f4979ce75150

PR #623, opened 2024-07-01
Make DAC capacity reporting conditional on whether DAC is in GDX

fschreyer:
Addresses #619 .


orichters (approved):
Thanks for fixing it that quickly.

fixes #619

PR #633, opened 2024-07-29
switch to US017 in reporting functions

fbenke-pik:


orichters:
I have no strong opinion on this, but [IIASA changes](https://github.com/IAMconsortium/legacy-definitions/pull/11/files) all units to `USD_2017`, not `US$2017`, mainly because they realized the $ sign created problems in some scripts, as far as I understood. We might use that as an opportunity to do that as well.

LaviniaBaumstark (approved):

R/reportEmployment.R
```diff
@@ -41,30 +41,30 @@ reportEmployment <- function(gdx, improvements, multiplier, subtype, shareManf,
...
-     "Tech|Electricity|Nuclear|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Solar|PV|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Solar|CSP|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Wind|Onshore|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Wind|Offshore|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Biomass|Combined Heat and Power w/o CC|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Oil|DOT|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Geothermal|OM Cost|fixed (US$2005/kW/yr)",
-     "Tech|Electricity|Storage|Battery|For PV|OM Cost|fixed (US$2005/kW/yr)",
+     "Tech|Electricity|Coal|Pulverised Coal w/o CC|Capital Costs (U$2017/kW)",
```
orichters, line 44: Here you typed `U$` instead of `US$`

fbenke-pik: Ah! Thanks for spotting

orichters:
What about
- https://github.com/fbenke-pik/remind2/blob/update_2024/vignettes/remind_summary.Rmd
- https://github.com/fbenke-pik/remind2/blob/update_2024/R/calcNetTradeValue.R#L11

By the way: What is the purpose of these `calc…` functions? Is anyone using them at all?

orichters (approved):
Looks good apart from these two tiny comments. Thanks for implementing that!

fbenke-pik:
> What about
> 
>     * https://github.com/fbenke-pik/remind2/blob/update_2024/vignettes/remind_summary.Rmd
> 
>     * https://github.com/fbenke-pik/remind2/blob/update_2024/R/calcNetTradeValue.R#L11
> 
> 
> By the way: What is the purpose of these `calc…` functions? Is anyone using them at all?

I think we can skip these for now. Not sure if the vignette and calc functions are still used, but we could remove stuff like this as part of a general cleanup of remind2 (there are more artifacts that could be removed).

Issue #628, opened 2024-07-15
cs2: add explicit historic variables


robertpietzcker:
With more and more versions of variables with slightly different definitions (w/ vs w/o bunkers, "LULUCF national accounting", w/ vs w/o non-energy use) it would be very helpful to have the possibility to add variables that should as well be displayed from "historical.mif". 

as an example, I added "Emi|GHG|w/ Bunkers|LULUCF national accounting" as a new variable from REMIND to cs2. 
Now as there is no variable with the same name in historical.mif, no historic lines appear.
it would be great if one could write something like this:
```
showLinePlots(data, "Emi|GHG|w/ Bunkers|LULUCF national accounting", historic="Emi|GHG|w/ Bunkers", modelsHistInclude="UNFCCC")
```

I think we will always have the situation where a source only provides one of the two versions, but we still want to be able to use it as comparison point also for the "slightly different" variable definition. 
Currently this is realized by just adding both names, eg:
```
items <- c(
  "Emi|GHG", # scenario
  "Emi|GHGtot") # historical
showLinePlots(data, items)
```

but this only works as long as REMIND doesn't provide both variables. Once you have versions such as "w/ Bunkers" where REMIND provides BOTH variables, adding both variable names would mean also both REMIND values are plotted in the same plot.

PR #635, opened 2024-07-30
refactor showLinePlots in Emissions notebook to use new parameters

fbenke-pik:
Choose historical variables to plot in Emission line plots explicitly.
See also https://github.com/pik-piam/remind2/issues/628 and https://github.com/pik-piam/mip/pull/108


robertpietzcker (approved):
looks good, thanks!

PR #636, opened 2024-08-06
avoid fail in case of NA, harmonize unit

orichters:
- if data contains `NA`, `any()` returns `NA` which cannot be evaluated by `if`
- harmonize unit to usual style


pweigmann (approved):
Thanks!

PR #639, opened 2024-08-08
lower memory footprint for "fill the demSeOth with zeroes so that oneRegi tests work"

lecfab:
Following @orichters' comment in https://github.com/pik-piam/remind2/pull/637, this PR reduces the memory footprint: v_demSeOth is now filled with zeroes up to a size 19x21x3 (the previous version filled it to 24000 instead of 3).

### DIFF
--- a/R/reportSE.R
+++ b/R/reportSE.R
@@ -372,8 +372,10 @@ reportSE <- function(gdx, regionSubsetList = NULL, t = c(seq(2005, 2060, 5), seq
   vm_demFeSector[is.na(vm_demFeSector)] <- 0
   # SE demand
   vm_demSe <- readGDX(gdx, "vm_demSe", field = "l", restore_zeros = F)[, y, ] * pm_conv_TWa_EJ
-  # SE demand of specific energy system technologies
-  v_demSeOth <- readGDX(gdx, c("v_demSeOth","vm_demSeOth"), field = "l", restore_zeros = T)[, y, ] * pm_conv_TWa_EJ
+  # SE demand of specific energy system technologies (ensure that all regions have a value)
+  v_demSeOth <- readGDX(gdx, c("v_demSeOth", "vm_demSeOth"), field = "l", restore_zeros = FALSE)[, y, ] * pm_conv_TWa_EJ
+  matrixRegionsYears <- new.magpie(cells_and_regions = getRegions(dataoc), years = y, fill = 0, sets = getSets(v_demSeOth, fulldim = FALSE))
+  v_demSeOth <- matchDim(v_demSeOth, matrixRegionsYears, dim=1, fill=0)  
   # conversion efficiency
   pm_eta_conv <- readGDX(gdx, "pm_eta_conv", field = "l", restore_zeros = F)[, y, ]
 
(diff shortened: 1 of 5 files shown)


orichters (approved):

R/reportSE.R
```diff
@@ -372,8 +372,10 @@ reportSE <- function(gdx, regionSubsetList = NULL, t = c(seq(2005, 2060, 5), seq
   vm_demFeSector[is.na(vm_demFeSector)] <- 0
   # SE demand
   vm_demSe <- readGDX(gdx, "vm_demSe", field = "l", restore_zeros = F)[, y, ] * pm_conv_TWa_EJ
-  # SE demand of specific energy system technologies
-  v_demSeOth <- readGDX(gdx, c("v_demSeOth","vm_demSeOth"), field = "l", restore_zeros = T)[, y, ] * pm_conv_TWa_EJ
+  # SE demand of specific energy system technologies (ensure that all regions have a value)
+  v_demSeOth <- readGDX(gdx, c("v_demSeOth", "vm_demSeOth"), field = "l", restore_zeros = FALSE)[, y, ] * pm_conv_TWa_EJ
+  matrixRegionsYears <- new.magpie(cells_and_regions = getRegions(dataoc), years = y, fill = 0, sets = getSets(v_demSeOth, fulldim = FALSE))
```
fbenke-pik, line 377: Not sure if it is necessary to create a dummy magclass object just for the matching here. Usually, the reference `ref` given to `matchDim` is the magclass object you which causes `mbind` to crash.

lecfab:
@fbenke-pik For reproducibility, the error happens with the gdx `/p/tmp/fabricel/powerRealizations/output/testOneRegi/fulldata.gdx`:

```
Error in mbind(tmp1, setNames(dimSums(mselect(v_demSeOth, all_enty = "seh2"),  : 
  Cannot handle objects! Spatial as well as data dimensions differ!
  Differences from first mbind() input:
  spatial:
      missing: `LAM`, `OAS`, `SSA`, `NEU`, ...
       having: `EUR`
  data:
      missing: `SE (EJ/yr)`, `SE|Biomass (EJ/yr)`, `SE|Electricity (EJ/yr)`, `SE|Electricity|Combined Heat and Power w/o CC (EJ/yr)`, ...
       adding: `SE|Input|Hydrogen|Other Energy System Consumption (EJ/yr)`
```

Error traceback is
1. convGDX2MIF(gdx)
2. convGDX2MIF.R#53: mbind(output, reportSE(gdx, regionSubsetList, t)[, t, ])
3. reportSE(gdx, regionSubsetList, t)
4. reportSE.R#401: mbind(tmp1, setNames(dimSums(mselect(v_demSeOth, all_enty = "seh2"), dim = 3), "SE|Input|Hydrogen|Other Energy System Consumption (EJ/yr)"))

R/reportSE.R
lecfab: Hey,
I tried to matchDim v_demSeOth with tmp1, but the dimension names don't match. That's why i used an intermediate named object.
What would be a more fitting solution?

lecfab: Falk's solution here: https://github.com/pik-piam/remind2/pull/644

PR #650, opened 2024-08-23
REMIND/EDGE-T variable harmonization

johannah-pik:
Change ES reporting unit notation from bn to billion as it is more compatible with other piam variables like billion US$2005/yr and currently imlemented in reporttransport.  This PR is related to [PR #361](https://github.com/pik-piam/piamInterfaces/pull/361) in piam_interfaces
 AND remove very old edge-transport reporting that was used ages ago and is not compatible with the new version


.buildlibrary
```diff
@@ -11,6 +11,7 @@ AcceptedNotes:
 - Imports includes .* non-default packages.
 - unable to verify current time
 - checking installed package size
+- no visible global function definition for ':='
```
orichters, line 14: Please rather add `@importFrom data.table :=` to `reportCrossVariables.R`, `plotNashConvergence.R` and `calc_CES_marginals.R` where it is currently missing.

And while you are at it, can you please adjust the unit of `Expenditure|Transport|Fuel` to `billion US$2005/yr` in [this line](https://github.com/pik-piam/remind/blob/master/R/reportCrossVariables.R#L239) such that we can close https://github.com/pik-piam/edgeTransport/issues/268?

orichters (approved):
Thanks for implementing that! Please adjust the two minor things.

.buildlibrary
johannah-pik: Sure! I did not know, that you can use it with dplyr as well..

orichters (approved):
Thanks!

PR #657, opened 2024-09-24
Add some missing emissions and cdr variables with national LULUCF accounting

fschreyer:
There is a land sink from existing forests accounted under national emissions reporting to the UNFCCC which is added to emissions variables including the specification ``LULUCF national accouting``. This PR completes the set of these variables for some missing CDR and emissions variables and reports the difference between Magpie and UNFCCC accounting as ``Emi|CO2|CDR|existing forest sink|LULUCF national accounting``. This is relevant for REMIND-EU runs which typically use the national LULUCF accounting.

This was discussed [here](https://mattermost.pik-potsdam.de/rd3/pl/6ga7yyhghib93e7fp4cs1nygqc).

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '231339003'
+ValidationKey: '231393712'
 AcceptedWarnings:
 - 'Warning: package ''.*'' was built under R version'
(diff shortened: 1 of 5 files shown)


fschreyer:
I need review on this.

fbenke-pik (approved):

Issue #658, opened 2024-09-26
check unit conversions in reportLCOE


orichters:
You are converting several times from 2005 to 2015, but I think that should be 2017 to 2015 now.

- https://github.com/pik-piam/remind2/blob/57c1fa01feabd4024808850466efcd06b4436af3/R/reportLCOE.R#L546
- https://github.com/pik-piam/remind2/blob/57c1fa01feabd4024808850466efcd06b4436af3/R/reportLCOE.R#L1027-L1028
- https://github.com/pik-piam/remind2/blob/57c1fa01feabd4024808850466efcd06b4436af3/R/reportLCOE.R#L1048-L1049
- https://github.com/pik-piam/remind2/blob/57c1fa01feabd4024808850466efcd06b4436af3/R/reportLCOE.R#L1215-L1216
- https://github.com/pik-piam/remind2/blob/57c1fa01feabd4024808850466efcd06b4436af3/R/reportLCOE.R#L1236-L1239
- https://github.com/pik-piam/remind2/blob/57c1fa01feabd4024808850466efcd06b4436af3/R/reportLCOE.R#L1538-L1556

PR #660, opened 2024-09-27
adjust unit conversion in reportLCOE

fbenke-pik:
This should fix https://github.com/pik-piam/remind2/issues/658


orichters (approved):
That looks very convincing, thanks for fixing it that quickly

PR #669, opened 2024-10-17
fix reportLCOE bugs

fbenke-pik:


lecfab (approved):
Thanks for doing that (Felix pointed it out but i thought it was not an issue because of the backward-compatibility interruption)

PR #672, opened 2024-10-23
skip test coverage in GitHub workflow

fbenke-pik:


.github/workflows/check.yaml
```diff
@@ -63,6 +63,6 @@ jobs:
         shell: Rscript {0}
         run: |
           nonDummyTests <- setdiff(list.files("./tests/testthat/"), c("test-dummy.R", "_snaps"))
-          if(length(nonDummyTests) > 0) covr::codecov(quiet = FALSE)
+          if(length(nonDummyTests) > 0 && !lucode2:::loadBuildLibraryConfig()[["skipCoverage"]]) covr::codecov(quiet = FALSE)
```
LaviniaBaumstark, line 66: do we need a special lucode version for this?

LaviniaBaumstark (approved):

.github/workflows/check.yaml
fbenke-pik: Yes, but lucode will require updating to the latest version anyways before running, so should be fine

PR #677, opened 2024-11-29
Add line plots for FE demand

robinhasse:
With this PR, we add the plots that @gunnar-pik requested in the validation meeting from 28/11/2024
- add line plots for sectoral FE demand at the beginning of each sector's line plots
  - removed IEA ETP reference data for buildings as there are much more recent references and the big legends squeezes the plot a lot
- add line plots for biomass and coal in section of `FE|Buildings|Solids`
- minor fix of headline hierarchy in industry section

### DIFF
--- a/inst/compareScenarios/cs_05_energy_demand.Rmd
+++ b/inst/compareScenarios/cs_05_energy_demand.Rmd
@@ -109,6 +109,11 @@ showAreaAndBarPlots(data, items2, tot, orderVars = "user", scales = "fixed")
 ```
 
 ### FE Buildings Line Plots
+```{r Buildings}
+showLinePlots(data, "FE|Buildings",
+              histModelsExclude = paste("IEA ETP", c("2DS", "B2DS", "RTS")))
+```
+
 #### Electricity
 ```{r Electricity}
 showLinePlots(data, "FE|Buildings|Electricity")
@@ -127,6 +132,8 @@ showLinePlots(data, "FE|Buildings|Gases")
 #### Solids
 ```{r Solids}
 showLinePlots(data, "FE|Buildings|Solids")
+showLinePlots(data, "FE|Buildings|Solids|Biomass")
+showLinePlots(data, "FE|Buildings|Solids|Fossil")
 ```
 
 #### Hydrogen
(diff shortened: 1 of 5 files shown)


inst/compareScenarios/cs_05_energy_demand.Rmd
```diff
@@ -191,7 +198,11 @@ showAreaAndBarPlots(data, items, tot, orderVars = "user", fill = TRUE)
 showAreaAndBarPlots(data, items2, tot, orderVars = "user", scales = "fixed")
 ```
 
-#### FE Industry Line Plots
+### FE Industry Line Plots
+```{r Industry}
+showLinePlots(data, "FE|Industry")
+```
```
ricardarosemann, line 204: I'm not sure if this still works, I don't find the industry line plots anymore.

inst/compareScenarios/cs_05_energy_demand.Rmd
```diff
@@ -127,6 +132,8 @@ showLinePlots(data, "FE|Buildings|Gases")
 #### Solids
 ```{r Solids}
 showLinePlots(data, "FE|Buildings|Solids")
+showLinePlots(data, "FE|Buildings|Solids|Biomass")
+showLinePlots(data, "FE|Buildings|Solids|Fossil")
```
ricardarosemann, line 136: These are computed in the post-processing, right? I'm wondering if there is a way to report in some way how reliable or not certain plots are. But this probably applies to many other plots as well.

ricardarosemann (approved):
Thanks for bringing back our plot for the totals.

I'm really sorry that it took me so long to review, I completely missed it.

inst/compareScenarios/cs_05_energy_demand.Rmd
robinhasse: Not sure I understand what you mean. The industry line plots follow below for each carrier under another subsection. But the line-plot headline was too low in the hierarchy. I corrected that and added the total line plot.

inst/compareScenarios/cs_05_energy_demand.Rmd
robinhasse: Yes, these are not native REMIND variable but computed [here](https://github.com/pik-piam/remind2/blob/ff306a764584ac6266a0cd4f2956335fe5713af4/R/reportFE.R#L968-L984) assuming that all traditional biomass is used in buildings. And only if there is more traditional biomass demand than biomass demand in buildings, there is a spill over into industry. But this should not happen anymore since remindmodel/remind#1519. I consider these variable as reliable as all other variables.

inst/compareScenarios/cs_05_energy_demand.Rmd
ricardarosemann: Maybe the reason that I couldn't find the line plots was that the header was too low in the hierarchy ... If you didn't have any problem when rebasing/merging then this is fine!

PR #686, opened 2025-01-27
clean up reportMacroEconomy

fbenke-pik:


orichters (approved):
thanks for fixing it that quickly. Seems that was really just a pointless line…

robertpietzcker:
thanks a lot, Falk!

PR #684, opened 2025-01-13
adapt emissions reporting to fixing and restructuring of industry feedstock equations

fschreyer:
These are the reporting changes required for https://github.com/remindmodel/remind/pull/1829.

PR #689, opened 2025-01-31
Ensure backwards compatibility for reportEmi

fbenke-pik:
Apply the Emission reporting before  https://github.com/pik-piam/remind2/pull/684 if necessary to ensure backwards-compatibility with the latest release.


fschreyer (approved):
Thanks for fixing this.

PR #692, opened 2025-02-13
Add a summary on per capita indicators

lecfab:
Add a section in compare scenario to gather per-capita affluence indicators.
Most of these graphs are already elsewhere, but it is hard to navigate through them when they are dispersed. 

[Example of compare scenario](https://cloud.pik-potsdam.de/index.php/apps/files/?dir=/Shared/ScenarioMIP/presentations&openfile=23393244)

image

image
image
image


dklein-pik:
Hi Fabrice, thank you for collecting the figures. Should the graphics perhaps be removed from the other places so that they do not appear twice in the already very large pdf?

dklein-pik (approved):

PR #697, opened 2025-02-27
remove building cs2 pdf from convGDX2mif test

fbenke-pik:
Removes running compareScenarios to create a dummy file from convGDX2mif test. This check adds little insight, but can cause hard to track and fix for Windows users who do not have the right Tex Version installed locally.


laurinks (approved):
Thank you, @fbenke-pik, for identifying and solving this problem!

PR #701, opened 2025-02-28
cs2: added 67th percentile plot to climate Section

robertsalzwedel:
## Purpose of this PR

This PR adds an additional plot to the climate section for the 67th percentile values. So far only the 50th percentile has been plotted explicitly. This is necessary as depending on the temperature target, either the 50th or 67th percentile is relevant. 

Not sure if there is a more elegant way to implement this? 

## Checklist:

- [x] cannot judge if there is more summation checks (as this is my first PR) but I do not see how an additional plot should change this

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '235123966'
+ValidationKey: '235316960'
 AcceptedWarnings:
 - .*following variables are expected in the piamInterfaces.*
 - Summation checks have revealed some gaps.*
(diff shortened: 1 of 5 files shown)


fbenke-pik (approved):

fbenke-pik:
>  cannot judge if there is more summation checks (as this is my first PR) but I do not see how an additional plot should change this

as you did not adjust any report functions, you did not introduce any new problems with the summation checks

PR #710, opened 2025-03-25
Add reportCrossSubmodels function

gabriel-abrahao:
## Purpose of this PR
This new function estimates extra non-CO2 emissions variables for sectors that REMIND doesn't directly represent non-CO2 emissions, but does represent activity levels. That includes for example N2O emissions from international shipping, for which EDGE-T tracks the activity with considerable detail, but not non-CO2 emissions, and CH4 emissions from the buildings sector. We use a simple approach, deriving emission factors from CEDS emissions and REMIND/EDGE-T endogenous 2020 activities.

Since some of these, especially the transport ones, require the EDGE-T reporting to be done and variables reported in vanilla REMIND, for now I believe it's best to keep it as a separate, post-processing function. We need these emissions with some urgency, but ultimately these EFs should probably be applied in each model's `report` functions. 

The resulting emission variables are labelled with `Extra`, for example `Emi|N2O|Extra|Transport|Bunkers|Freight`. That's there to indicate that these emissions are currently **not** accounted for in the totals. So, for example, `Emi|N2O|Extra|Transport|Bunkers|Freight` is simply not part of `Emi|N2O`. We need a bigger discussion on how to handle these summations. Adding those to the totals would mean that `reportEmi` is not self-contained anymore, and could lead to even more confusion. 

So, as we need these for ScenarioMIP, my proposal is to keep those emissions as `|Extra|` for now, add this function to REMIND's `reporting.R` script, and then start a conversation on the best way to include them in the totals.

## Checklist:
- [x] do not create new complaints about summation checks. **No complaints but is definitely worrisome for the summations, see above**


R/reportCrossSubmodels.R
```diff
@@ -0,0 +1,148 @@
...
+#' @export
+#' @importFrom assertr assert not_na
+#' @importFrom gdx readGDX
+#' @importFrom magclass getYears getRegions mbind setNames mselect
+#' new.magpie setYears mcalc dimReduce
+#' @importFrom tibble as_tibble
+#' @importFrom tidyselect everything
+#' @importFrom tidyr everything
+#' @importFrom madrat readSource toolGetMapping toolAggregate
+#' @importFrom mrcommons everything
```
gabriel-abrahao, line 19: Hi RSE people ( @LaviniaBaumstark @fbenke-pik ), this function needs a `read` function from `mrcommons`, but that's not a requirement for `remind2`. So this actually doesn't work in a clean environment, just in one where `mrcommons` is already loaded. What would be the procedure in a case like this?

fbenke-pik: You are not allowed to use any madrat functions in the reporting. Anything you want to use must come through data in the gdxes of the REMIND run or auxiliary files. This is the same problem discusses here: https://github.com/remindmodel/remind/issues/1969#issuecomment-2626994117

gabriel-abrahao: Oh, yes, sorry, I remember that discussion. Indeed we should find a more elegant way to do this. In this case now, it's more of a quickfix due to ScenarioMIP, but a long-term solution here would involve adding more complexity to REMIND.

gabriel-abrahao: For now @fbenke-pik @LaviniaBaumstark , would you say it's OK to commit a 1.4Mb file to `remind2`? It's about the size of the current `inst` folder. I get that this is not ideal though. One alternative would be to put this function *in* REMIND itself instead, for use in the reporting.

fbenke-pik: No, it is not ok to add a 1.4Mb file to the remind2 lib. But you won't need it either. The data you are actually using should not be more than 20kb. 

For now, I propose the following approach:

1. Write a calc function im mrremind like this

```
calcFoo <- function() {
  cedsVars <- c(
    "1A4a_Commercial-institutional",
    "1A4b_Residential",
    "1A3ai_International-aviation",
    "1A3aii_Domestic-aviation",
    "1A3di_International-shipping",
    "1A3di_Oil_Tanker_Loading"
  )
  ceds <- readSource("CEDS2024")[, 2020, c("ch4", "n2o_n")][, , cedsVars]

  return(list(x = ceds, weight = NULL, unit = "foo", description = "bar"))
}
```

2. Add a call in fullREMIND.R under "files used in reporting"

`calcOutput("Foo", file = "foo.cs4r")`

This is just for documentation in mrremind so we know which files are currently generated solely to use them for additional reporting.

3. Generate the data with resolution of 21 regions, so all regions you might possibly need in REMIND are covered.
4. Add the file to the remind2 repo and use it in your reporting function. 

For an example of this approach, see "Other Fossil in Electricity"
https://github.com/pik-piam/remind2/blob/master/R/reportCrossVariables.R#L399
https://github.com/pik-piam/mrremind/blob/master/R/fullREMIND.R#L233

R/reportCrossSubmodels.R
```diff
@@ -0,0 +1,148 @@
...
+#' @importFrom piamutils deletePlus
+#'
+
+reportCrossSubmodels <- function(inmifpath) {
```
fbenke-pik, line 23: In which context do you want to execute this function? If it is part of convGDX2MIF, it should rather use the gdx as input. If it should run outside of convGDX2MIF (which I assume is the case), it could be a function here or a script in REMIND. I am not sure what is better, need more context for the execution of this function.

fbenke-pik:
@gabriel-abrahao Closing for now, as we opted for another approach not including remind2

PR #717, opened 2025-04-07
Update CS2 - add many transport plots, and specify what is with and without bunkers

robertpietzcker:
## Purpose of this PR

Improve the usefulness of cs2. 
I added many transport plots, and specified which emission variables are with and without bunkers

## Checklist:


inst/compareScenarios/cs_03_emissions.Rmd
```diff
@@ -2,8 +2,29 @@
...
+```{r GHG total w/o Bunkers}
+showLinePlots(data, "Emi|GHG|w/o Bunkers", histVars = "Emi|GHG")
+```
+
+### GHG total w/ Bunkers
+```{r GHG total w/ Bunkers}
+showLinePlots(data, "Emi|GHG|w/ Bunkers")
+```
+
+### GHG - total
```
lecfab, line 27: Shouldn't "Emi|GHG" be redundant with either "Emi|GHG|w/ Bunkers" or "Emi|GHG|w/o Bunkers"?

inst/compareScenarios/cs_03_emissions.Rmd
```diff
@@ -127,11 +161,16 @@ showLinePlots(
 )
 ```
 
+### Energy without bunkers
```
lecfab, line 164: For consistency, i would put "without bunkers" or "w/o Bunkers" in all the titles.
My preference goes to the full word, but fine with any choice of abbreviation and case!

lecfab (approved):
Thanks Robert that's a lot of useful information added!
My comments are minor so I "approve" in case you need to merge quickly.

Could you upload the pdf or give a path? it's more intuitive to understand the changes by directly going through the pdf.

inst/compareScenarios/cs_03_emissions.Rmd
robertpietzcker: mostly. 
The old "Emi|GHG" was a mix - it was "w/o bunkers" for regions and "w/ bunkers" for global.

In the mid-term, I would propose to remove the plot for "Emi|GHG", but in the old mifs (Remind3.4 and earlier) we don't have the full explicit "w/ " and "w/o" reporting, so for the next half year/year when people want to compare to the older version, I left it in. (my attempts of rerunning the new reporting with the old runs failed. I think there were some parameters added that are missing in the old gdx, causing the reporting to fail)

PR #725, opened 2025-04-10
Code-cleaning: Improve Carbon Management|Storage variable calculation

amerfort:
## Purpose of this PR
This PR is for code cleaning. `Carbon Management|Storage` variables are now calculated all at once by selecting the respective `Carbon Management|Carbon Capture` variable and multiplying the share of geologically stored carbon from captured carbon.
As the level of detail on Industry subsector carbon management is not necessary, we exclude this level of detail from the generic calculation.

## Checklist:


tabeado:
Wow, thanks for this great simplification, @amerfort! I really appreciate the effort to shorten the code.
My intuition would be to give "include" and "exclude" names that are more specific to the calculation they are used in, e.g. CarMaVar_include/exclude or something like that. But that's more for @fbenke-pik to judge.

fbenke-pik:
> My intuition would be to give "include" and "exclude" names that are more specific to the calculation they are used in, e.g. CarMaVar_include/exclude or something like that

That is a bit subjective in this case. Basically a tradeoff between conciseness/readability and how explicit/specific you want to be.

fbenke-pik (approved):

amerfort:
As Felix and I are the responsible people for the emissions reporting, I will change the "include" "exclude" naming to a more meaningful thing as soon as we need it to avoid ambiguity. Thanks for the point Tabea, I'll keep it in mind but merge it as is for now :)

PR #733, opened 2025-05-28
Fix waste CCU calculation in emissions reporting for EU21 for some edge cases

fschreyer:
## Purpose of this PR

Addresses parts of these [summation errors](https://github.com/remindmodel/development_issues/issues/545) (only emi summation errors in DEU remaining, which is a different bug). Not using p_share_CCS in the waste CCU calculation led to some NAs and zeros down the line. 

## Checklist:

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '238793235'
+ValidationKey: '238825272'
 AcceptedWarnings:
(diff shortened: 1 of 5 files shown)


amerfort (approved):
Wow, thanks for fixing this. Just for my understanding: the previous calculation lead to summation errors in EU21 runs, because the respective numbers of EU regions were too small and hence rounding errors occured?

PR #735, opened 2025-06-04
ienhance reportCrossVariables to deal with missing curtailment variables

fbenke-pik:
## Checklist:


fbenke-pik:
@irr-github When adding your new realization, you will have to decide whether your new realization should be added here or not https://github.com/pik-piam/remind2/blob/master/R/reportSE.R#L74-L82

@fschreyer Let me know if this is the right way to tackle the Problem, or if we should address this bug differently.

PR #736, opened 2025-06-04
 remove no longer needed condition in reportSE

fbenke-pik:
## Purpose of this PR

remove check for power module realization, that is no longer needed with exactly one realization (and assuming, that all future realizations include storage loss)


irr-github:
thanks for the fix

PR #746, opened 2025-07-25
Fixing "Fuel costs for own ESM" reporting for secondary energy trade scenarios

Renato-Rodrigues:
## Purpose of this PR

- Including secondary energy trade import and export costs not considered in the "Fuel costs for own ESM" variable
- Replace NAs by zeros in "Res|Average Supply Costs"

## Checklist:


R/reportPrices.R
```diff
@@ -48,6 +48,8 @@ reportPrices <- function(gdx, output = NULL, regionSubsetList = NULL,
     output <- mbind(output, reportExtraction(gdx, regionSubsetList = regionSubsetList, t = t))
     message("- reportMacroEconomy")
     output <- mbind(output, reportMacroEconomy(gdx, regionSubsetList = regionSubsetList, t = t)[, getYears(output), ])
+    message("- reportMacroEconomy")
```
robertpietzcker, line 51: shouldn't this message read "reportTrade"?

inst/compareScenarios/cs_02_macro.Rmd
```diff
@@ -203,15 +203,53 @@ vars <- intersect(paste0("Price|Final Energy|Transport|", carriers, "|Rawdata"),
...
+### PE Trade
+
+```{r PE Trade}
 vars <-
   levels(data$variable) %>%
   str_subset("^Trade\\|(Uranium.)?[^|]*$")
 walk(vars, showLinePlots, data = data)
 ```
 
+### SE Imports Trade
```
robertpietzcker, line 235: was there a specific reason you decided to plot imports and exports separately? 
for PE, we only show the net trade, and I think that is sufficient information also for SE in the compScen, as there should be no region that both imports and exports SE

robertpietzcker (approved):

R/reportTrade.R
```diff
@@ -88,8 +88,6 @@ reportTrade <- function(gdx,regionSubsetList=NULL,t=c(seq(2005,2060,5),seq(2070,
   tmp <- mbind(tmp,setNames(Xport[,,"perm"] * C_2_CO2 * 1000,                                  "Trade|Exports|Emi Allowances|Volume (Mt CO2-equiv/yr)"))
 
   # SE Trade
-  if (module2realisation["trade",2] == "se_trade") {
```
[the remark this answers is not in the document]
Renato-Rodrigues: I tested with NPi and it just fills the variable with all zeros. 
If we keep the conditional here, we will need to do conditionals in all SE trade code and in the compariosn scenario because these variables are necessary for aggregating other variables and for displaying cs results.

PR #747, opened 2025-07-29
deal with floating point comparison in test_ranges

fbenke-pik:
## Purpose of this PR

Deal with floating point comparison in test_ranges so that it no longer produces false positives as happened here `/p/projects/remind/modeltests/remind/output/SSP2-EU21-NPi-AMT_2025-07-25_22.18.56`

## Checklist:

### DIFF
--- a/R/test_ranges.R
+++ b/R/test_ranges.R
@@ -56,8 +56,10 @@ test_ranges <- function(data, tests, reaction = c('message', 'stop'),
          'must be a string.')
   }
 
-  # Test all variables in data_variables agains low and up
+  # Test all variables in data_variables against low and up
   .test <- function(data_variables, low, up) {
+
+    tolerance <- 1e-10
 
     # get the name of the variable dimension in the magpie object
     variable_name <- data_variables %>%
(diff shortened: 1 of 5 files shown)


R/test_ranges.R
```diff
@@ -68,7 +70,7 @@ test_ranges <- function(data, tests, reaction = c('message', 'stop'),
     low_data <- if (!is.null(low) && any(data_variables < low)) {
       data_variables %>%
         magclass_to_tibble() %>%
-        filter(.data$value < low) %>%
+        filter(.data$value < low - tolerance) %>%
```
pweigmann, line 73: If my understanding is correct, the tolerance should be added instead of subtracted, no? In an example with the lower threshold being 0 we would want to add a tolerance of the lower threshold being positive 1^-10 instead I think?

R/test_ranges.R
```diff
@@ -80,7 +82,7 @@ test_ranges <- function(data, tests, reaction = c('message', 'stop'),
     up_data <- if (!is.null(up) && any(data_variables > up)) {
       data_variables %>%
         magclass_to_tibble() %>%
-        filter(.data$value > up) %>%
+        filter(.data$value > up + tolerance) %>%
```
pweigmann, line 85: Same as above

R/test_ranges.R
pweigmann: Makes sense, I got it the wrong way around!

pweigmann (approved):

PR #784, opened 2026-03-12
Add DACCS, ERW and OAE to Revenue|Government|Tax|Carbon|non-ES CDR

amerfort:
## Purpose of this PR
This PR adds the expenditure (negative tax revenues) that the regulator would need to spend on DACCS, ERW or OAE. These emissions were previously missing from the revenue calculation and are now accounted for with this new variable (component) Revenue|Government|Tax|Carbon|non-ES CDR.

## Checklist:

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -16,3 +16,4 @@
 ^.*CITATION.cff$
 pull_request_template.md
 ^_pkgdown\.yml$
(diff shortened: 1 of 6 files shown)


JiaruiZhongPik (approved):
Thanks for the implementation

PR #794, opened 2026-05-09
handle module reanming in reportLCOE

fbenke-pik:
## Checklist:

### DIFF
--- a/R/reportLCOE.R
+++ b/R/reportLCOE.R
@@ -65,6 +65,14 @@ reportLCOE <- function(gdx, output.type = "both") {
 
   # get module realizations
   module2realisation <- readGDX(gdx, "module2realisation")
+
+  ## Ensure backwards compatibility for release version 3.6.0 (can be removed with 3.7.0)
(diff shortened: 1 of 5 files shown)


R/reportLCOE.R
```diff
@@ -65,6 +65,14 @@ reportLCOE <- function(gdx, output.type = "both") {
 
   # get module realizations
   module2realisation <- readGDX(gdx, "module2realisation")
+
+  ## Ensure backwards compatibility for release version 3.6.0 (can be removed with 3.7.0)
+  if ("CCU" %in% module2realisation$modules) {
+    CCU_mod <- module2realisation[module2realisation$modules == "CCU", 2]
+  } else {
+    CCU_mod <- module2realisation[module2realisation$modules == "carbonUtilization", 2]
```
ricardarosemann: Very minor: I'd say that this is rather the realisation than the module, so a different name (`CCU_real` or similar instead of `CCU_mod`) might be more intuitive. But really not important ...

ricardarosemann (approved):
Thank you for the quick fix!

PR #796, opened 2026-05-21
Include only ccsinje* technologies in the calculation of the geological storage potential 

dklein-pik:
## Purpose of this PR

Adjust `reportEmi()` to not include irrelevant technologies in the calculation of the geological storage potential by explicitly selecting CCS injection technologies and a specific grade in `max_geolStorage`. In a few cases, the upper bound of `vm_co2ccs` contains values not only for ccsinje but also for other technologies, which can be ignored.

## Checklist:

### DIFF
--- a/man/remind2-package.Rd
+++ b/man/remind2-package.Rd
@@ -20,7 +20,6 @@ Useful links:
 
 Authors:
 \itemize{
-  \item Renato Rodrigues \email{renato.rodrigues@pik-potsdam.de}
   \item Lavinia Baumstark
   \item Falk Benke
   \item David Bantje
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '41542548'
+ValidationKey: '41579286'
 AcceptedWarnings:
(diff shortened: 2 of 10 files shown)


amerfort:
Thanks for solving this so swiftly. In the long run we shall make sure that there are no false entries in upper limits of non-relevant technologies.

man/remind2-package.Rd
```diff
@@ -20,7 +20,6 @@ Useful links:
 
 Authors:
 \itemize{
-  \item Renato Rodrigues \email{renato.rodrigues@pik-potsdam.de}
```
amerfort, line 23: I don't understand why this line should be deleted?

amerfort (approved):
Thanks David! 

I cannot judge the changes in `man`, so maybe, if necessary, get additional feedback from other people? 
But for `reportEmi.R`, this now looks good as a hot fix.
I'll open an issue to fix the excessive dimensionality of vm_co2CCS in EU21 scenarios and another one where we can discuss whether it is necessary to remove the rlf dimension from vm_co2CCS (the cause of this issue in the first place).

PR #797, opened 2026-05-26
Update CompareScenarios to show cumulated CO2 emissions consistent with budget formulation in REMIND

laurinks:
## Purpose of this PR

Update CompareScenarios to show cumulated CO2 emissions consistent with budget formulation in REMIND:

- In `remind2::reportEmi()`, cumulated emissions from 2005 are computed as follows based on [remind2::time_cumulate()](https://github.com/pik-piam/remind2/blob/master/R/time_cumulate.R): Emissions are linearly interpolated and then summed from 2005 to the respective year. For example, cumulated emissions from 2005 to 2025 are the sum of yearly emissions from 2005, 2006, ..., 2025, i.e. 21 time steps.
- To correctly show cumulated emissions from 2020 in CompareScenarios, the computation has been corrected to `cum.value.period - cum.value.2020 + yearly.value.2020` so that the 2020 value is included.

Assumptions on the exclusion of start year / end year in the definition of cumulated emissions depends on the convention. Here, we choose the approach that is consistent with the [REMIND internal computation](https://github.com/remindmodel/remind/blob/5af5c7a6270523ce87703eb805b14f9ab4a6a65d/core/postsolve.gms#L11-L20).

image

Notes that lines have been switched to 650, 750 and 1000 in line with AMTs. Test before this change shown above and here: `/p/tmp/laurinko/compScen`

FYI: Only changes in `inst/compareScenarios/cs_03_emissions.Rmd` were manually implemented, other changes from `buildLibrary()`.

## Checklist:

### DIFF
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '4325790'
+ValidationKey: '4346389'
 AcceptedWarnings:
 - .*following variables are expected in the piamInterfaces.*
 - Summation checks have revealed some gaps.*
--- a/.buildlibrary
+++ b/.buildlibrary
@@ -1,4 +1,4 @@
-ValidationKey: '41579286'
+ValidationKey: '4325790'
 AcceptedWarnings:
 - .*following variables are expected in the piamInterfaces.*
 - Summation checks have revealed some gaps.*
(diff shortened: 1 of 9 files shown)


lecfab (approved):
Thanks for adding the offset!

PR #801, opened 2026-06-05
actual variable to check EUR emission targets

lea-hayez:
## Purpose of this PR
Adjust variable plotting in compScen. The actual variable to check EUR emission targets is: "Emi|GHG|w/ Intra-region Bunkers|LULUCF national accounting". Note that UK is treated as having the same 55% GHG emission reduction target as EU27 in the plot, although the real UK target is stricter (~68%, https://www.theccc.org.uk/climate-action/ ).

## Checklist:


lea-hayez:
@fbenke-pik how do I know if this PR is creating "new" complaints? 

When running buildlibrary, I get:

**1. Summation check**
```
# Run summation check on a total of 4632 variables.
variable groups found in additional_summation_checks.csv
Emi|CO2|Cumulated >  
...
The child sum differs by 0.0362% to 0.0362% from the parent, absolute difference up to 0.0216 Mt CO2.
```

**2. Reporting check**
A long list of around 50 variables for:
```
4. Test if REMIND reporting is produced as it should and check data integrity ('test-convGDX2mif.R:40:7') - The following variables are expected in the piamInterfaces package for templates AR6, AR6_NGFS, ELEVATE, NAVIGATE, SHAPE, ARIADNE, ECEMF, ScenarioMIP,
but cannot be found in the reporting generated by /p/tmp/R_tmp/RtmpBVJlDw/fulldata-SSP2-NPi-AMT.gdx:
```

Both seem unrelated to my changes, I hope!

robertpietzcker (changes requested):
Could you please also change the historic variable? 
I don't think ""historical" = "Emi|GHGtot" # historical" will be a good comparison. 

I guess you actually need to look into the historical.mif and look at all the available UNFCCC GHG emission variables to determine which one is the one you want (and if you don't find the exact one, at least the one that has the closest values in 1990 and 2022 to what you expect)

robertpietzcker:
> Note that UK which should be excluded to check if target is reached

I don't understand that sentence :-)

fbenke-pik:
Looking at your PR, you did not any complaints produced by buildLibrary, as you only adjusted a plot in the comparison documents and did not touch any reporting logic.

So this is good to go on my end.

fbenke-pik (approved):

fbenke-pik:
However, I cannot comment on whether the plot compares the right things and I also don't understand the sentence regarding the UK.

lea-hayez:
Thanks @robertpietzcker ! The historical UNFCCC "Emi|GHG|LULUCF national accounting" should be close enough.

RE UK: I was only pointing out the target of 55% is not applied to the UK, but I am fine with keeping as is.

robertpietzcker:
> RE UK: I was only pointing out the target of 55% is not applied to the UK, but I am fine with keeping as is.

just to prevent that I am confused again when reading it - could you change the sentence in the description to something like 
"Note that UK is treated as having the same 55% GHG emission reduction target as EU27 in the plot, although the real UK target is stricter (~68%, https://www.theccc.org.uk/climate-action/ )"

:-)

Thanks!

robertpietzcker:
> Thanks @robertpietzcker ! The historical UNFCCC "Emi|GHG|LULUCF national accounting" should be close enough.

sorry for being picky - could you quickly change to 
"Emi|GHG|w/o Bunkers|LULUCF national accounting" instead? 

I didn't think clearly when asking you to determine which of the available UNFCCC GHG variables you want - given that intra-EU bunkers in REMIND is calculated by taking 1/3 of bunkers, using the "w/ bunkers" variant is 2/3 of bunkers off, while the "w/o bunkers" is only 1/3 of bunkers off....

if it is a large hassle now, this can also happen in some later update of cs2, I just thought I'd raise it.

robertpietzcker (approved):

PR #807, opened 2026-07-02
Adding CO2 price convergence plots to nashConvergenceReport

gabriel-abrahao:
## Purpose of this PR
Adds a new template to `nashConvergenceReport`, with basic information about the CO2 price convergence algorithm when `carbonprice = functionalForm`. 

## Checklist:

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: remind2
 Title: The REMIND R package (2nd generation)
-Version: 2.2.2
-Date: 2026-06-24
+Version: 2.3.0
+Date: 2026-07-06
 Authors@R: c(
     person("Renato", "Rodrigues", , "renato.rodrigues@pik-potsdam.de", role = c("aut", "cre")),
     person("Lavinia", "Baumstark", role = "aut"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -66,7 +66,6 @@ Imports:
     piamPlotComparison (>= 0.0.10),
     piamutils,
     plotly (>= 4.10.4),
-    qpdf,
     quitte (>= 0.3132.0),
     readr,
     remulator,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -66,6 +66,7 @@ Imports:
     piamPlotComparison (>= 0.0.10),
     piamutils,
     plotly (>= 4.10.4),
+    qpdf,
     quitte (>= 0.3132.0),
     readr,
     remulator,
--- a/R/reportFE.R
+++ b/R/reportFE.R
@@ -19,10 +19,8 @@
 #'
 #' @export
 #' @importFrom gdx readGDX
-#' @importFrom magclass new.magpie mselect getRegions getYears mbind setNames
-#'                      getNames<- as.data.frame as.magpie getSets
-#' @importFrom dplyr %>% filter full_join group_by left_join mutate rename
-#'     select semi_join summarize ungroup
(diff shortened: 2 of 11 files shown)


gabriel-abrahao:
@fbenke-pik `buildLibrary` keeps failing with:

```
❯ checking for unstated dependencies in examples ... OK
   WARNING
  ‘qpdf’ is needed for checks on size reduction of PDFs
```

Which I don't really understand where is coming from. Adding the `qpdf` dependency to `DESCRIPTION` doesn't fix it either. I do have `qpdf` available for `library(qpdf`)`. Any tips?

laurinks (commented):
Thank you, @gabriel-abrahao! Good initiative! Could you add an example of the outcome to make it easier to grasp?

DESCRIPTION
```diff
@@ -66,6 +66,7 @@ Imports:
     piamPlotComparison (>= 0.0.10),
     piamutils,
     plotly (>= 4.10.4),
+    qpdf,
```
fbenke-pik: why do we need a new dependency for this? should be avoided at any cost, if you can achieve the same with existing libraries.

R/reportPrices.R
```diff
@@ -26,8 +26,8 @@
 #'
 #' @importFrom dplyr %>% case_when distinct filter inner_join tibble left_join rename
 #' @importFrom gdx readGDX
-#' @importFrom magclass mbind getYears getRegions setNames dimExists new.magpie
-#' lowpass complete_magpie getItems<- getNames unitsplit unitjoin
+#' @importFrom magclass mbind getYears getRegions setNames dimExists new.magpie lowpass
```
fbenke-pik, line 29: i assume you did that because of the new warning `"@importFrom must be only 1 line long, not 2". ` 

it was imposed by the latest roxygen updates, but should not stop from buildLibrary from succeeding. if this is indeed the case, please just go back to the old from, as multiple `importFrom` for the same library don't seem like a good idea to me

R/reportFE.R
```diff
@@ -19,10 +19,8 @@
 #'
 #' @export
 #' @importFrom gdx readGDX
-#' @importFrom magclass new.magpie mselect getRegions getYears mbind setNames
-#'                      getNames<- as.data.frame as.magpie getSets
-#' @importFrom dplyr %>% filter full_join group_by left_join mutate rename
-#'     select semi_join summarize ungroup
+#' @importFrom magclass new.magpie mselect getRegions getYears mbind setNames getNames<- as.data.frame as.magpie getSets
```
fbenke-pik, line 22: see my comment on reportPrices

fbenke-pik:
You should be able to fix the `qpdf` problem by adding an Accepted Warning. Should work by adaption what's done in piamValidation: https://github.com/pik-piam/piamValidation/blob/main/.buildlibrary#L6

R/reportPrices.R
fbenke-pik: Update: do as you wish. You can either ignore the warning, as it does not stop buildLibrary from succeeding or keep it as it is.

DESCRIPTION
fbenke-pik: see my other comment on how to fix this

gabriel-abrahao:
Thanks for the tips @fbenke-pik , worked like a charm! I also removed the unnecessary dependency. @laurinks here's an example. We'll probably need to have REMIND write some other parameters to the GDX to make this more useful, but it's a start. You'll see more of it on the group meeting today :)

[nashConvergence-co2price.html](https://github.com/user-attachments/files/29696939/nashConvergence-co2price.html)

fbenke-pik (approved):

PR #810, opened 2026-07-06
Adding reasons for non-convergence to nashConvergenceReport

gabriel-abrahao:
## Purpose of this PR
DISCLAIMER: The code in this PR was, in its vast majority, done by our new AI friend Claude. I have thoroughly reviewed the logic and had a good look at all the code, but there are for example some functions (like `setTimeLimit`) that I'm not well acquainted with, but seemed to do the job in all the tests I ran. Do we have some guidelines, or at least best practices, on AI-majority PRs?

This extends the current `nashConvergence-overview` template to include a section that automatically parses `full.lst` for the `Reasons for non-convergence` displays, and gathers them in a slidable widget per iteration that you can use along with all the other `nashConvergence` plots to diagnose differerent convergence issues. It's particularly useful when the reasons are not covered by existing plots (such as carbon price convergence criteria), and gives a good overview of what happens at each iteration.

The function that does this relies on the assumption that there is a `full.lst` in the same folder as the `fulldata.gdx`, and fails gracefully when there isn't one. By default, it relies on `grep` and `sed` to parse the file quickly (~10s), but in absence of those tools it falls back to a `read_lines` implementation that takes significantly longer (~60s). To avoid mysterious issues in the future if `full.lst` gets even larger, there's also an implemented 90s time limit for the parsing. if it goes above this time it will simply say so in the report and shouldn't affect anything else. 

Example output:

Full file:
[nashConvergence-overview.html](https://github.com/user-attachments/files/29704746/nashConvergence-overview.html)

Screenshot:

image

## Checklist:


fbenke-pik:
> Do we have some guidelines, or at least best practices, on AI-majority PRs?

I won't have time to review before next week, but can already name you some rules to follow:
- quality of the code should be the same as for human written code
- you should fully understand what's going on
- reviewing your code should not take the reviewer longer than for regular PRs
- no AI slop

fbenke-pik:
> Do we have some guidelines, or at least best practices, on AI-majority PRs?

Pik-wide: https://gitlab.pik-potsdam.de/sdmc/ai-workflow/-/wikis/home/ai-guidelines
Preliminary rules by RSE: https://gitlab.pik-potsdam.de/REMIND/wiki/-/wikis/Preliminary-REMIND-RSE-AI-guidelines

gabriel-abrahao:
@fbenke-pik I think this is following all of these. I also fixed the merge conflicts, accepting this version here that was higher

R/toolParseLSTReasons.R
```diff
@@ -0,0 +1,200 @@
...
+#'   \code{NULL}. Default 90.
+#' @return A \code{data.frame} with integer column \code{iteration} and character
+#'   column \code{text}, one row per iteration ordered by iteration, or
+#'   \code{NULL} if \code{lstFile} does not exist, the time limit is exceeded, or
+#'   parsing otherwise fails.
+#'
+#' @author Gabriel Abrahão
+#'
+#' @export
+toolParseLSTReasons <- function(lstFile, timeLimit = 90) {
```
fbenke-pik: as `tool` has a special meaning in our toolchain (it is used for a specific type of madrat functions), please avoid using it in other contexts. I suggest `parseLstReasons`

R/parseLSTReasons.R
```diff
@@ -0,0 +1,200 @@
...
+      ends <- cumsum(lens)
+      starts <- ends - lens + 1L
+      idx <- which(hasLines)
+      for (j in seq_along(idx)) {
+        texts[idx[j]] <- paste(out[starts[j]:ends[j]], collapse = "\n")
+      }
+    }
+
+    data.frame(iteration = blocks$iteration, text = texts, stringsAsFactors = FALSE)
+  }, error = function(e) NULL)
```
fbenke-pik, line 172: Please add more sophisticated error handling or drop this form of error handling.

As of now, errors are swallowed silently, which would lead to this simply not being executed at all without anyone noticing. However, this could easily break if the underlying messages in the lst file change and I would prefer this to crash then so someone notices and fixes it over the code lying around dormant for years.

fbenke-pik (changes requested):

fbenke-pik:
I had a  look at the code, but would like to defer the final review for this to @Renato-Rodrigues , as he knows the architecture of the nash convergence report best.

fbenke-pik:
Is this new feature useful enough to be added here?

How is the newly parsed section different from the info in the "Convergence Criterias" section? Does it provide additional info that cannot be found in other plots or is it just a convenience feature presenting the same info in a different way? 

Also, wouldn't it be easier to learn the right `less` commands on the cluster to parse the lst file by hand and take a look as opposed to generating this html file for the same info?

Just asking, because  I don't do this type of analyses often, but I'd like to hear the opinion of @LaviniaBaumstark, @robertpietzcker  and @Renato-Rodrigues.

gabriel-abrahao:
> Is this new feature useful enough to be added here?
> 
> How is the newly parsed section different from the info in the "Convergence Criterias" section? Does it provide additional info that cannot be found in other plots or is it just a convenience feature presenting the same info in a different way? 

This prints the all the actual reasons for non convergence as REMIND understood in the section that handles non convergence. It was useful for example in diagnosing a case where the conditional structure had a bugged and it was interpreting a technically converged run as non converged. 

It also mitigates some level of obsolescence of the scripts here. For example, the carbon price path convergence was included more rigorously in the last year, and it was not covered at all by this report. Reporting the actual REMIND stated problems makes it a lot easier to diagnose things that are not yet covered. 

> 
> Also, wouldn't it be easier to learn the right `less` commands on the cluster to parse the lst file by hand and take a look as opposed to generating this html file for the same info?

Well, a similar argument could be made for a lot of the information in these reports, but having them summarised and comparable without jumping around huge LSTs and GDXs has proved a big time (and once life) saver before, at least to me. 

Besides, it's pretty hard to keep track of exact iterations when navigating the LST, and it would take a pretty complex bash command to get rid of the less useful statements in between. In fact, such a bash command is exactly what this is doing in "fast" mode. 

My bottom line here is that this is a helper for debugging complex non convergence reasons. So it's not essential, but I had one instance already since this PR where I don't think I would have found the bug without it so soon. I've been debugging it on and off for a couple of weeks without success, and looking at this output per iteration paired with tbe carbon price convergence report made the problem immediately clear.

robertpietzcker:
I think it is helpful to have this information in one place and easy to digest - while I myself am fine with looking at the full.lst, I think it is a hurdle for newcomers. So the html may help some/many REMIND users
 
My one "complaint" with the html would be that the upper graphical part does not contain all the information, so it would be great if that could be added in the future :-)

R/parseLSTReasons.R
gabriel-abrahao: The idea was for it to not really fail silently, but for the markdown to handle the errors [here](https://github.com/gabriel-abrahao/remind2/blob/4fda3da38d11fa7976a4d655df806ace78b24e35/inst/markdown/nashConvergence-overview.Rmd#L73) instead of that happening in the function. This allows it to not stop the execution of others part of the reporting, and for the issue to simply be shown in that section instead of the desired output, making it non-threatening to others uses of the reporting. 

If the function fails and returns NULL or a length zero vector, that text div will show:

In case the lenght is zero (which could be due to something changing in the structure):
`"*No per-iteration non-convergence sections found in `full.lst`.*"`

If the `full.lst` file wasn't found where expected:
`"*No `full.lst` found next to the gdx (expected at `", lstFile,"`) - per-iteration diagnostics unavailable.*"`

If the function returns NULL even though the file exists:
`"*Parsing `full.lst` exceeded the 90s time limit or failed; per-iteration diagnostics unavailable.*"`

I believe this strikes a balance between future-proofing against this function not working anymore, having useful output regardless of it failing, and letting the user know that something should have happened but actually failed.

Does this address your concerns?

R/toolParseLSTReasons.R
gabriel-abrahao: Done

gabriel-abrahao:
Also we`ll need your force merge powers again @fbenke-pik , checks are complaining about a node.js version

Renato-Rodrigues:
I will probably have time to look into this next week or the week after.

I also have a parallel PR that is not finished yet, which should update the convergence criteria analysis and hopefully cover the missing information and Gabriel's case as well.

The Nash convergence report was based on two principles: keeping dependencies to a minimum, and focusing on easy-to-understand graphical representations of the convergence criteria and related information.

If possible, I would like to preserve these principles. Adding a wall of text or a dependency on `full.lst`, even if optional, would go against the original design.

If there is no urgency to merge this, I would ask to hold off on the merge until I have had a chance to take a look at it.

fbenke-pik:
> Also we`ll need your force merge powers again @fbenke-pik , checks are complaining about a node.js version

RIght now, the PR says "All checks have passed", so  I assume this has resolved itself.

R/parseLSTReasons.R
fbenke-pik: Am I talking to Gabriel or to Claude right now? :D ( I really hope you did not simply paste a Claude response to my concerns here.)

I think the error handling should be part of the function, not the markdown calling it and there are ways to make sure it fails gracefully within the function as well. Separating error throwing and handling errors like this is bad practice in my eyes, cause every calling entity has to duplicate the same error handling and the markdown has a lot of function logic in it.

However, as Renato has raised more fundamental concerns about this PR, let's not further discuss this right now, as this might be obsolete.

gabriel-abrahao: Hehehe, it indeed looks like an AI response, but I actually did not use AI for that reply at all! I suppose the style is seeping into our human consciousness now, which is concerning in itself... Rest assured I will never send any AI-generated anything without disclaiming it first, and that goes double for text with my name beside it.

That makes sense, but let's indeed wait for Renato's review and I can rework the error handling if he gives his OK

robertpietzcker:
> Adding a wall of text or a dependency on full.lst, even if optional, would go against the original design.

hm. 
It seems to me that Gabriel is actually using the html, which is cool,  and is putting effort into developping it further to make it even more useful for him, so I would say "does the change make the html a lot larger, or make it much more likely to fail? if not, I would say great that someone uses it and develops it further!". 

as no REMIND run will ever happen without creating a full.lst, I don't think a dependence on that file is problematic.

Renato-Rodrigues:
closed in favor of consolidation at: https://github.com/pik-piam/remind2/pull/819

PR #811, opened 2026-07-13
Move remulator dependency to Suggests

johanneskoch94:
This should remove the hard-dependency on the terra package.


johanneskoch94:
The question @dklein-pik is whether the function `readSupplycurveBio` is even used - or if it can be deleted.

dklein-pik:
> The question @dklein-pik is whether the function `readSupplycurveBio` is even used - or if it can be deleted.

Yes, it is used in two REMIND output scripts:
`/output/single/emulator.R` and
`/output/comparison/supplycurves.R`.

dklein-pik (approved):
Great if this avoids the terra dependency! Thanks!

johanneskoch94:
Yeah, this should remove the hard-dependencies on magpie4 and thus terra :) 
I somehow missed the output scripts! Thanks for pointing them out - I won't delete the function then, but would it perhaps make sense to move it to remulator?

PR #812, opened 2026-07-14
Drop direct dependency on reshape2

johanneskoch94:
- Drop direct dependency on discontinued reshape2 package
- Bump version

reshape2 has long been discontinued.


fbenke-pik (approved):
Thank you for this!

